PR #50, opened 2020-08-16
Add a streaming version of list_objects 

notoriaga:
Was messing around with something and needed to iterate over a lot of keys. This lets you do work on the batches as they come in. Also a fun intro to streams!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -538,29 +538,23 @@ where
 
     match res {
         Ok(hoo) => {
-            if let Some(delete_marker) = hoo.delete_marker {
-                eprintln!("delete_marker: {}", delete_marker);
-                if delete_marker {
-                    return Ok(None);
-                }
-            }
-            if let Some(e_tag) = hoo.e_tag {
-                return Ok(Some(e_tag));
+            if let Some(true) = hoo.delete_marker {
+                Ok(None)
+            } else if let Some(e_tag) = hoo.e_tag {
+                Ok(Some(e_tag))
+            } else {
+                Err(anyhow!("head_object failed (3): No e_tag found: {:?}", hoo))
             }
         }
         Err(RusotoError::Unknown(e)) => {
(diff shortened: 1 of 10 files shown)


notoriaga:
Also made `S3Obj` public, because I want the etags as well and that saves a lot of requests. Not sure if that's alright

src/lib.rs
```diff
@@ -388,23 +390,61 @@ where
...
-                bucket_contents.push(entry.key);
+                entry.key
+            })
+        })
+        .collect();
+
+    Ok(keys)
+}
+
+pub fn s3_stream_objects<'a, T>(
```
silverjam, line 408: Should this be something like `s3_list_objects_stream`?  Which seems a bit verbose... so maybe we can combine with renaming the non-streaming version too... so

`s3_list_objects` -> `s3_list` (add a backward compat wrapper for the old name)
`s3_stream_objects` -> `s3_list_streaming`

If you know of a better convention for this, feel free to use that.

notoriaga: Yeah I think that makes sense. Another way to shorten it could be to also remove `s3_` from all the functions, it does seem a little redundant. Although that would be a breaking change so maybe not worth it atm

silverjam: @notoriaga I'm fine with a breaking change if we add backward compat wrappers, or just rev the major version

notoriaga: I removed all the `s3_` prefixes and I think it looks good. There was a name conflict then with some non-pub functions so those got renamed

```
head_object -> head_object_request
list_objects -> list_objects_request
```

`_request` meaning those are the functions at actually call the s3 api. `s3_stream_objects` is now `list_objects_stream`, which is a little wordy but also descriptive

notoriaga: Also one small implementation change, the sync down function now uses `list_objects_stream` so we don't have to duplicated the logic for paginating the list request

PR #71, opened 2020-10-16
Sync Between Buckets - [ROOST-110]

betog25:
I'm open to other naming for the sync functions, if others are more ideal.  Currently, I just exposed the `sync_remote_to_local ` and `sync_local_to_remote `

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -900,6 +900,7 @@ fn process_globs<'a>(
 ) -> Option<&'a str> {
     let mut excluded = false;
     let mut included = false;
+
     for pattern in glob_excludes {
         if pattern.matches(path) {
             excluded = true;
@@ -1124,35 +1125,28 @@ where
     let mut stream = list_objects_stream(s3, source_bucket, source_prefix);
 
     while let Some(from_entries) = stream.try_next().await? {
-        // let entries_to_copy = from_entries.iter().filter(should_sync_copy);
-        // let futs = entries_to_copy.map(|e| copy_object_request(...))
-        // try_join_all(futs).await?;
-
         for entry in from_entries {
             if let S3ListingItem::S3Object(obj) = entry {
-                if obj.key.ends_with(".txt") {
-                    let object_info = copy_object_request(
-                        s3,
-                        source_bucket,
-                        source_prefix,
-                        &obj.key,
-                        dest_bucket,
-                        dest_prefix.unwrap(),
-                    )
-                    .await?;
+                let path = process_globs(&obj.key, &glob_includes, &glob_excludes);
+                match path {
+                    Some(y) => {
+                        let object_info = copy_object_request(
+                            s3,
+                            source_bucket,
+                            source_prefix,
+                            &obj.key,
+                            dest_bucket,
+                            dest_prefix.unwrap(),
+                        )
+                        .await?;
+                    }
+                    None => (),
                 }
             }
         }
     }
 
     Ok(())
-}
-
-async fn should_sync_copy(params: &CopyObjectParams) -> bool {
-    // filename matches globs
-    // and
-    // file in dest does not exist or etags don't match
-    true
 }
 
 #[cfg(test)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -402,23 +402,6 @@ impl std::fmt::Debug for SyncParam {
     }
 }
 
-impl std::str::FromStr for SyncParam {
-    type Err = String;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if s.starts_with("s3://") {
-            Ok(SyncParam::Bucket {
-                bucket: "todo".to_owned(),
-                path: "todo".to_owned(),
-            })
-        } else {
-            Ok(SyncParam::Local {
-                path: "todo".to_owned(),
-            })
-        }
-    }
-}
-
 #[logfn(err = "ERROR")]
 pub async fn sync<T>(
     s3: &T,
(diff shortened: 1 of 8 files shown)


silverjam (changes requested):
Not sure I'm a fan of the API change yet since it's not an incremental change, and I'm not understanding exactly why this is better.

I think I would prefer adding a new direction which could be something like "bucket", "bucket-to-bucket", "cross-bucket", or "internal" then name the new sync function cold be "sync_buckets".

At any rate, if we go with the API rename, we'll need to fix this wrapper script: https://github.com/swift-nav/esthri/blob/master/bin/aws.esthri because it uses the "sync direction" concept.

betog25:
@silverjam Makes sense.  I can revert that part and go with a third direction, with `up` and `down`

silverjam:
> @silverjam Makes sense. I can revert that part and go with a third direction, with `up` and `down`

Using `sideways` would work too 😄

betog25:
> > @silverjam Makes sense. I can revert that part and go with a third direction, with `up` and `down`
> 
> New direction named `sideways` would work too 😄

Are we sure we want to add a third direction, versus creating a function like `sync-across`?  Adding a third direction to `sync` would require all other `sync` calls to supply a destination bucket

silverjam:
> > > @silverjam Makes sense. I can revert that part and go with a third direction, with `up` and `down`
> > 
> > 
> > New direction named `sideways` would work too 😄
> 
> Are we sure we want to add a third direction, versus creating a function like `sync-across`? Adding a third direction to `sync` would require all other `sync` calls to supply a destination bucket

We can discuss this more "offline", but maybe the API change we should go for is one `sync` call that allows the direction of the sync to be inferred from the input parameters.  If we changed the input type of the `sync` function from simple string to an enum of "Bucket" or "Local" we could eliminate the "direction" enum and then infer the direction based on the input params.  That is, something like:

```rust
enum SyncParam {
  Local { path: String },
  Bucket { bucket: String, path: String },
}
```

silverjam:
Looking good, can we update https://github.com/swift-nav/esthri/blob/master/bin/aws.esthri as well?

src/lib.rs
```diff
@@ -391,24 +391,28 @@ fn bytes_range(offset: u64) -> String {
     format!("bytes={}-", offset)
 }
 
+pub enum SyncParam {
+    Local { path: String },
+    Bucket { bucket: String, path: String },
+}
+
+impl std::fmt::Debug for SyncParam {
```
silverjam, line 399: Just use `#[derive(Debug)]` on the type, no need to implement this manually unless you're customizing something.

src/lib.rs
```diff
@@ -391,24 +393,44 @@ fn bytes_range(offset: u64) -> String {
...
+pub enum SyncParam {
+    Local { path: String },
+    Bucket { bucket: String, path: String },
+}
+
+impl std::str::FromStr for SyncParam {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s3_format = Regex::new(r"^s3://+?(?P<bucket>[a-zA-Z0-9._-]+)/(?P<prefix>.*)$").unwrap();
```
silverjam, line 406: For our purposes I think we can match "anything that's not a `/`" for the bucket name.

src/main.rs
```diff
@@ -85,17 +84,10 @@ enum Command {
...
         #[structopt(long)]
-        bucket: String,
-        #[structopt(long)]
-        key: String,
-        /// The directory to use for up/down sync
+        source: String,
         #[structopt(long)]
-        directory: String,
-        /// Optional include glob pattern (see man 3 glob)
+        destination: String,
```
silverjam, line 90: Should be possible for the type here to be SyncParam, structopt should automatically pick up the FromStr instance.

tests/integration/sync_test.rs
```diff
@@ -93,15 +90,15 @@ fn test_sync_up() {
...
-    let res = blocking::sync(
-        s3client.as_ref(),
-        SyncDirection::up,
-        crate::TEST_BUCKET,
-        &s3_key,
-        &local_directory,
-        &includes,
-        &excludes,
-    );
+    let source = SyncParam::Local {
```
silverjam, line 93: Let's add constructors for Local and Bucket to get rid of some of this boilerplate

src/lib.rs
```diff
@@ -399,11 +399,20 @@ pub enum SyncParam {
...
+        SyncParam::Bucket { path, bucket }
+    }
+}
+
 impl std::str::FromStr for SyncParam {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let s3_format = Regex::new(r"^s3://+?(?P<bucket>[a-zA-Z0-9._-]+)/(?P<prefix>.*)$").unwrap();
+        let s3_format = Regex::new(r"^s3://+?(?P<bucket>[^/]+)/(?P<prefix>.*)$").unwrap();
```
silverjam, line 415: playing around with the regex here: https://regex101.com/r/vZL9L9/1 -- I don't think the `+?` after `s3://` is necessary

tests/integration/sync_test.rs
```diff
@@ -14,8 +14,8 @@ fn test_sync_down() {
     let includes: Option<Vec<String>> = Some(vec!["*.txt".to_string()]);
     let excludes: Option<Vec<String>> = Some(vec!["*".to_string()]);
 
-    let source = SyncParam::new_bucket(crate::TEST_BUCKET.to_owned(), s3_key.to_owned());
-    let destination = SyncParam::new_local(local_directory.to_owned());
+    let source = SyncParam::new_bucket(&crate::TEST_BUCKET.to_owned(), &s3_key.to_owned());
```
silverjam, line 17: Sometimes you can't just take the linter output verbatim, since we're trying to get rid of the `.to_owned()` calls here, you can just simply pass `crate::TEST_BUCKET` since it's already a `&str`, same with `s3_key` here.  If the type is a `String`, then pass `&var_name` and remove the `.to_owned()`.

silverjam (approved):

notoriaga (approved):

PR #77, opened 2020-10-27
Setup graceful handler for SIGINT/SIGTERM [SITL-62]

silverjam:
Use warp "graceful shutdown" features so that the SIGTERM received from kubernetes will initiate a clean shutdown of connections.  We'll still need to configure timeouts in kubernetes to something longer than the default (60 seconds), the graceful shutdown will not allow new connections, but will allow existing downloads/browsing to continue until they complete.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -162,7 +162,7 @@ async fn main() -> Result<()> {
 
     match cli.cmd {
         Put { bucket, key, file } => {
-            setup_upload_ctrlc_handler();
+            setup_upload_termination_handler();
             upload(&s3, &bucket, &key, &file).await?;
         }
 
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -158,12 +158,11 @@ async fn main() -> Result<()> {
     let credentials_provider = DefaultCredentialsProvider::new().unwrap();
     let s3 = S3Client::new_with(http_client, credentials_provider, Region::default());
 
-    setup_cancel_handler();
-
     use Command::*;
 
     match cli.cmd {
         Put { bucket, key, file } => {
+            setup_upload_ctrlc_handler();
```
betog25, line 165: I'm just curious what the `ctrlc` stands for/means.  Controller?

betog25: Nevermind, I found this https://docs.rs/ctrlc/3.1.7/ctrlc/

betog25 (approved):

src/main.rs
silverjam: fixed

silverjam:
@notoriaga merging, if you have any comments I can fix in another PR

PR #97, opened 2020-11-17
Enable building against rustls [SITL-223]

silverjam:
+ Add support for building Rusoto with rustls: https://github.com/ctz/rustls -- this allows static linking of esthri without having to create a musl build of openssl

+ Makes building the CLI optional, which is helpful for projects like SITL that use esthri as a library

+ Adds more Makefile.toml tasks to handle the juggling of the rustls/nativels features

+ The binary build of `esthri` now create a fully static binary

+ Rename's `esthri_lib` crate to just `esthri`


src/esthri/Cargo.toml
```diff
@@ -0,0 +1,121 @@
...
+flate2 = "1.0"
+tempdir = "0.3"
+backtrace = "0.3"
+
+[lib]
+name = "esthri"
+path = "src/lib.rs"
+
+[features]
+default = ["ctrlc"]
```
notoriaga, line 43: Does this need to be default if building the cli is optional?

src/esthri/Cargo.toml
```diff
@@ -0,0 +1,121 @@
...
+name = "esthri"
+version = "4.1.3"
+authors = [ "Jason Mobarak <jason@swift-nav.com>",]
+edition = "2018"
+
+[dependencies]
+once_cell = "1.5"
+rust-crypto = "0.2"
+hex = "0.4"
+structopt = "0.3"
```
notoriaga, line 16: Is there a way we can move the binary only dependencies to optional as well? It might make sense to break the cli and lib into two crates if we want to minimize time to build when consuming as a lib

notoriaga (approved):

PR #165, opened 2021-05-12
remove tail, make public API params more generic

silverjam:
prep work for eventual release to crates.io:
- generalize parameters more, use AsRef<_> to make parameter types more flexible
- switch back to anyhow instead of eyre
- remove custom wrapper modules around rusoto
- make blocking module optional
- remove flawed tail command


src/esthri/src/blocking.rs
```diff
@@ -19,69 +20,91 @@ use super::Result;
 use super::SyncParam;
 
 #[tokio::main]
-pub async fn head_object<T>(s3: &T, bucket: &str, key: &str) -> Result<Option<ObjectInfo>>
+pub async fn head_object<T, SR0, SR1>(s3: &T, bucket: SR0, key: SR1) -> Result<Option<ObjectInfo>>
 where
     T: S3 + Send,
+    SR0: AsRef<str>,
```
silverjam, line 26: Is this the best way to do this?  Would like to be able to pass a &str and String for either or both of the bucket/key param.... using one type parameter seems to require that they both be the same concrete type.

src/esthri/src/http_server.rs
```diff
@@ -48,7 +48,7 @@ use async_stream::stream;
 
 use futures::stream::StreamExt;
 
-use eyre::{eyre, Report};
+use anyhow::anyhow;
```
silverjam, line 51: Switch back to anyhow, it supports backtraces in stable now too and eyre didn't get us the backtraces that we hoped it would (maybe user error?)

src/esthri/src/lib.rs
```diff
@@ -23,18 +23,21 @@ use std::marker::Unpin;
...
 use log::*;
 use log_derive::logfn;
 use once_cell::sync::Lazy;
-use tokio::{io::AsyncReadExt, time};
+use tokio::io::AsyncReadExt;
 use walkdir::WalkDir;
 
+pub use anyhow::Result;
+
+#[cfg(feature = "blocking")]
```
silverjam, line 40: make the blocking module optional

src/esthri/src/main.rs
```diff
@@ -52,15 +59,6 @@ enum Command {
         /// The path of the local file to write
         file: String,
     },
-    /// Tail an object from S3
-    Tail {
```
silverjam, line 56: Remove tail command since it didn't really work

src/esthri/Cargo.toml
```diff
@@ -38,26 +37,46 @@ name = "esthri"
...
-default = []
+default = [ "rustls", "cli", "http_server", "blocking", ]
 aggressive_lint = []
+blocking = []
 http_server = [ "async-tar", "async-compression", "async-stream", "bytes", "futures-util", "mime_guess", "sluice", "sanitize-filename", "serde", "serde_derive", "tokio-util", "warp", "ctrlc",]
-nativetls = [ "rusoto_nativetls",]
-rustls = [ "rusoto_rustls",]
-cli = [ "ctrlc", "structopt",]
+nativetls = [ "rusoto_credential", "rusoto_core/native-tls", "rusoto_signature", "rusoto_s3/native-tls", "hyper-tls",]
+rustls = [ "rusoto_credential", "rusoto_core/rustls", "rusoto_signature", "rusoto_s3/rustls", "hyper-rustls",]
```
silverjam, line 45: we don't need our own wrapper modules around rusoto to enable features since we can turn on features with the `<dep>/<feature_name>` syntax

src/esthri/src/blocking.rs
notoriaga: Yup I believe that's correct

src/esthri/src/http_server.rs
notoriaga: What about going to thiserror/a hand rolled error enum - anyhow is more for applications. As a consumer of esthri as a library it would be nice to have a fixed set of errors to work with

notoriaga: Probably a decent amount of work though

notoriaga:
Also we should all +1 this - https://github.com/awslabs/aws-sdk-rust/issues/16

src/esthri/src/http_server.rs
silverjam: I think it's worth doing

silverjam: Done

silverjam:
> 
> 
> Also we should all 👍 this - [awslabs/aws-sdk-rust#16](https://github.com/awslabs/aws-sdk-rust/issues/16). Knowing aws it'll still probably be complicated enough to use to warrant esthri 😄

upvoted

src/esthri/src/http_server.rs
silverjam: the optional `http_server` module still uses anyhow, but I removed `anyhow` from the main lib module

src/esthri/src/lib.rs
```diff
@@ -759,27 +741,39 @@ async fn head_object_request<T>(s3: &T, bucket: &str, key: &str) -> Result<Optio
...
 
-    match res {
-        Ok(hoo) => process_head_obj_resp(hoo),
-        Err(RusotoError::Unknown(e)) => {
+    match res.as_mut() {
+        Some(Ok(_)) => {
+            let hoo = res.unwrap().unwrap();
+            process_head_obj_resp(hoo)
+        }
+        Some(Err(RusotoError::Unknown(e))) => {
```
silverjam, line 762: This is a bit nuts, in order to deal with the borrow checker, I need to wrap this in an `Option` so that I would borrow `e` (a sub-component of the `res` object) then move that object later when constructing the wrapped error.  I thought I would need to use `Option::take` but it turned out to be unnecessary, not sure why.

Makefile.toml
```diff
@@ -20,98 +20,38 @@ TLS_PROVIDER = "nativetls"
...
 
 [tasks.build]
 command = "cargo"
 args = [
-	"build",
-	"--features=${TLS_PROVIDER},http_server,cli",
-	"@@remove-empty(BUILD_FLAVOR)",
-	"@@remove-empty(TARGET)"
-	]
+        "build",
```
john-michaelburke, line 33: What tool did you use to format this Makefile.toml?

src/esthri/src/lib.rs
```diff
@@ -549,41 +519,48 @@ where
...
+    bucket: SR0,
+    key: SR1,
+) -> impl TryStream<Ok = Vec<S3ListingItem>, Error = Error> + Unpin + 'a
 where
     T: S3 + Send,
+    SR0: AsRef<str> + 'a,
+    SR1: AsRef<str> + 'a,
 {
-    list_objects_stream_with_delim(s3, bucket, key, None)
+    list_objects_stream_with_delim(s3, bucket, key, None as Option<&str>)
```
john-michaelburke, line 532: Maybe you could create a variable with a name representative of the parameter here.

john-michaelburke (approved):

Makefile.toml
silverjam: This is hand formatted. but it would be nice if there was as toml "prettifier" tool out there, maybe there is?

john-michaelburke: Might already be built in to cargo-make actually.
https://github.com/sagiegurari/cargo-make/blob/dc907c6e9ce41209115a4ee60db8b70f299f84fb/src/lib/descriptor/makefiles/toml.toml#L5

silverjam: 🎉

src/esthri/src/lib.rs
silverjam: Done

PR #172, opened 2021-05-27
concurrent sync support [SITL-248]

silverjam:
Added tests for a few bugs (introduced while writing this (and the previous) PR, but not caught by unit tests):
- nested directory sync
- downloading a zero length file

Baseline aws sync remote to local:
```
$ time bash -c "aws s3 sync s3://drive-test-logs-2/2021-04/2021-04-27-BayLoop/Ublx_M8L_GLO_ADR431_SBAS_1Hz_AmotechL2 d &>sync_d.log"

real    0m11.524s
user    0m4.426s
sys     0m2.996s
```

New code  (remote to local):
```
$ time bash -c "cargo run --release -- sync --source s3://drive-test-logs-2/2021-04/2021-04-27-BayLoop/Ublx_M8L_GLO_ADR431_SBAS_1Hz_AmotechL2/ --destination d2/ &>sync_d2.log"

real    0m6.510s
user    0m2.961s
sys     0m2.378s
```

Baseline local to remote sync:
```
$ time bash -c "aws s3 sync d s3://drive-test-logs-2/d3 &>sync_d3.log"

real    0m5.412s
user    0m3.248s
sys     0m1.513s
```

New code (local to remote):
```
$ time bash -c "cargo run --release -- sync --source d2/ --destination s3://drive-test-logs-2/d4/ &>sync_d4.log"

real    0m6.116s
user    0m1.729s
sys     0m1.240s
```

Basic validation:
```
$ du -s d
226818  d

$ du -s d2
226818  d2

$ aws s3 sync s3://drive-test-logs-2/d4 d4 &>/dev/null
$ aws s3 sync s3://drive-test-logs-2/d3 d3 &>/dev/null

$ du -s d3 d4
226818  d3
226818  d4
```

### DIFF
--- a/src/esthri/Cargo.toml
+++ b/src/esthri/Cargo.toml
@@ -37,13 +37,13 @@ name = "esthri"
 path = "src/lib.rs"
 
 [features]
-default = [ "rustls", "cli", "http_server", "blocking", ]
+default = [ "rustls", "cli", "http_server", "blocking", "tokio",]
 aggressive_lint = []
 blocking = []
 http_server = [ "async-tar", "async-compression", "async-stream", "bytes", "futures-util", "mime_guess", "sluice", "sanitize-filename", "serde", "serde_derive", "tokio-util", "warp", "ctrlc", "anyhow"]
 nativetls = [ "rusoto_credential", "rusoto_core/native-tls", "rusoto_signature", "rusoto_s3/native-tls", "hyper-tls",]
 rustls = [ "rusoto_credential", "rusoto_core/rustls", "rusoto_signature", "rusoto_s3/rustls", "hyper-rustls",]
-cli = [ "ctrlc", "structopt", "blocking" ]
+cli = [ "ctrlc", "structopt", "tokio/rt", "tokio/rt-multi-thread",]
 
 [dependencies.anyhow]
 version = "1"
@@ -85,7 +85,8 @@ optional = true
 
 [dependencies.tokio]
 version = "1.6"
-features = [ "io-std", "net", "rt", "time", "rt-multi-thread"]
+features = [ "io-std", "net", "time",]
+optional = true
 
 [dependencies.backoff]
 version = "0.3"
--- a/src/esthri/Cargo.toml
+++ b/src/esthri/Cargo.toml
@@ -24,6 +24,7 @@ thiserror = "1.0"
 chrono = "0.4"
 regex = "1"
 envy = "0.4"
+async-stream = "0.3"
 
 [dev-dependencies]
 md5 = "0.7"
@@ -37,10 +38,10 @@ name = "esthri"
 path = "src/lib.rs"
 
 [features]
-default = [ "rustls", "cli", "http_server", "blocking", "tokio",]
+default = [ "rustls", "cli", "http_server", "blocking",]
 aggressive_lint = []
 blocking = []
-http_server = [ "async-tar", "async-compression", "async-stream", "bytes", "futures-util", "mime_guess", "sluice", "sanitize-filename", "serde", "serde_derive", "tokio-util", "warp", "ctrlc", "anyhow"]
+http_server = [ "async-tar", "async-compression", "bytes", "futures-util", "mime_guess", "sluice", "sanitize-filename", "serde", "serde_derive", "tokio-util", "warp", "ctrlc", "anyhow"]
 nativetls = [ "rusoto_credential", "rusoto_core/native-tls", "rusoto_signature", "rusoto_s3/native-tls", "hyper-tls",]
 rustls = [ "rusoto_credential", "rusoto_core/rustls", "rusoto_signature", "rusoto_s3/rustls", "hyper-rustls",]
 cli = [ "ctrlc", "structopt", "blocking", "tokio/rt-multi-thread",]
@@ -85,8 +86,7 @@ optional = true
 
 [dependencies.tokio]
 version = "1.6"
-features = [ "io-std", "net", "time", "rt",]
-optional = true
+features = [ "io-std", "rt",]
 
 [dependencies.backoff]
 version = "0.3"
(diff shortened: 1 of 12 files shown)


src/esthri/Cargo.toml
```diff
@@ -84,8 +84,8 @@ version = "0.3"
 optional = true
 
 [dependencies.tokio]
-version = "1.5"
-features = [ "io-std", "net", "rt", "time",]
+version = "1.6"
+features = [ "io-std", "net", "rt", "time", "rt-multi-thread"]
```
silverjam, line 88: Maybe the `"rt-multi-thread"` feature should be optional?

notoriaga: I think ideally the rt and rt-multi-threaded features should be optional/only enabled for the cli. I don’t think esthri as a library needs a run time, just the AsyncWrite/AsyncRead traits

silverjam: @notoriaga looks like we need the "rt" feature at least since I'm using `tokio::task::spawn_blocking`

src/esthri/src/lib.rs
```diff
@@ -1084,54 +1094,104 @@ where
...
+    directory: &'a Path,
+    glob_includes: &'a [Pattern],
+    glob_excludes: &'a [Pattern],
+) -> impl Stream<Item = Result<(String, Option<ListingMetadata>)>> + 'a {
+    async_stream::stream! {
+        for entry in WalkDir::new(directory) {
+            let entry = if let Ok(entry) = entry {
+                entry
+            } else {
+                yield Err(entry.err().unwrap().into());
```
john-michaelburke, line 1107: Would it make more sense here and below to use a return? I feel like using the break implies other things will happen after breaking from the loop.

src/esthri/src/lib.rs
```diff
@@ -1084,54 +1094,104 @@ where
...
+            let stripped_path = path.strip_prefix(&directory);
             let stripped_path = match stripped_path {
                 Err(e) => {
                     warn!("unexpected: failed to strip prefix: {}", e);
-                    continue;
+                    return Ok(());
                 }
                 Ok(result) => result,
             };
             let stripped_path = format!("{}", stripped_path.display());
```
john-michaelburke, line 1184: I know this was not in your PR but I stumbled on a neat way for converting a Path/PathBuf to a Cow using ".to_lossy_string()" and borrowing this gives you an &str.

src/esthri/src/lib.rs
silverjam: Changed

john-michaelburke (approved):

src/esthri/src/lib.rs
silverjam: Thanks, cleaned up what I could find

src/esthri/Cargo.toml
```diff
@@ -37,13 +37,13 @@ name = "esthri"
 path = "src/lib.rs"
 
 [features]
-default = [ "rustls", "cli", "http_server", "blocking", ]
+default = [ "rustls", "cli", "http_server", "blocking", "tokio",]
```
notoriaga, line 40: I think serde and async-stream also need to be marked as non-optional. I think originally they were only used for the server but they are both used in lib.rs now

src/esthri/Cargo.toml
```diff
@@ -84,8 +84,9 @@ version = "0.3"
 optional = true
 
 [dependencies.tokio]
-version = "1.5"
-features = [ "io-std", "net", "rt", "time",]
+version = "1.6"
+features = [ "io-std", "net", "time", "rt",]
```
notoriaga, line 88: Could `net` be moved to `http_server`? also I don't see `tokio::time` anywhere so maybe `time` can be removed

src/esthri/Cargo.toml
silverjam: 🛠️

src/esthri/Cargo.toml
silverjam: Was able to remove net and time, I think we use `tokio::net` directly (probably do via warp though)

src/esthri/Cargo.toml
notoriaga: Ahh I see, because the etag stuff is blocking? Probably could be replaced with tokio fs but what you're doing is pretty much how that works afaik

notoriaga:
With `cargo check --features rustls --no-default-features` (should be how sitl uses this) I get

```
(base) {10:23}~/projects/esthri:silverjam/parallel-sync ✓ ➭ cargo check --features rustls --no-default-features
    Checking esthri v5.1.2 (/home/steve/projects/esthri/src/esthri)
warning: field is never read: `bucket`
  --> src/esthri/src/types.rs:22:5
   |
22 |     pub(crate) bucket: Option<String>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `key`
  --> src/esthri/src/types.rs:23:5
   |
23 |     pub(crate) key: Option<String>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: field is never read: `upload_id`
  --> src/esthri/src/types.rs:24:5
   |
24 |     pub(crate) upload_id: Option<String>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: 3 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 1.62s
```

which is strange because the global data is reference in `upload_from_reader` which is a public function, not quite sure whats up

silverjam:
> With `cargo check --features rustls --no-default-features` (should be how sitl uses this) I get
> [..]
> which is strange because the global data is referenced in `upload_from_reader` which is a public function, not quite sure whats up

Strange, using `pub(super)` instead of `pub(crate)` seemed to fix this issue

notoriaga (approved):

PR #174, opened 2021-06-03
transparent download/upload (de)compression [DEVINFRA-491]

silverjam:
Adds "transparent (de)compression" for upload and download, two new API functions are added:
- `esthri::upload_compressed`
- `esthri::download_decompressed`

For upload, the file is simply compressed in serial prior to upload, this works will with concurrent uploading so that we can easily know what the final size of the file is.

For download, it was a little more complex, decompression is weaved into the download process so a new trait was created and the download "strategy" was abstracted into a trait.

Misc: switched to `impl AsRef<..>` for a lot of the parameters since it removed the weirdly named type parameters like `SR0` and `SR1` that needed to be unique in case a caller was using different types for each parameter.


src/esthri/src/ops/download.rs
```diff
@@ -0,0 +1,557 @@
...
+*
+* This source is subject to the license found in the file 'LICENSE' which must
+* be be distributed together with this source. All other rights reserved.
+*
+* THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+* EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+* WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+*/
+
+use std::fs::File;
```
notoriaga, line 13: Using stuff from `std::io` and `std::fs` together with async might result in some weird behavior because they have potential to block. I guess we've been doing this all along and haven't really run into problems so maybe not a big deal. It might be because we tend to use the multi-threaded runtime so blocking one thread doesn't block everything. Regardless, using  `tokio:fs`/wrapping the blocking stuff in `spawn_blocking` would probably be a good idea.

src/esthri/Cargo.toml
```diff
@@ -40,9 +40,10 @@ name = "esthri"
 path = "src/lib.rs"
 
 [features]
-default = [ "rustls", "cli", "http_server", "blocking",]
+default = [ "rustls", "cli", "http_server", "blocking", "compression",]
```
john-michaelburke, line 43: Do the rust powers that be have any guidelines on the use of hyphens versus underscores for feature names? Just looking at this file it seems a few of them use either/or.

src/esthri/src/main.rs
```diff
@@ -44,6 +44,9 @@ struct Cli {
 enum Command {
     /// Upload an object to S3
     Put {
+        /// Should the file be compressed during upload
+        #[structopt(long)]
+        compress: bool,
         #[structopt(long)]
         bucket: String,
```
john-michaelburke, line 51: If the comments here are what make it to the cli usage/help maybe it would be worth adding the format the s3 bucket should be in? e.g.
"drive-logs2/path/to/log"
or
"s3://drive-logs2/path/to/log"

src/esthri/tests/integration/upload_test.rs
```diff
@@ -16,6 +17,28 @@ fn test_upload() {
     assert!(res.is_ok());
 }
 
+#[test]
+fn test_upload_compressed() {
```
john-michaelburke, line 21: nit. This compression functionality may be a decent candidate for a benchmark or two. I think out of scope for this PR though.

john-michaelburke (approved):

src/esthri/Cargo.toml
silverjam: I think I've seen both, so I don't think there's a consensus

src/esthri/tests/integration/upload_test.rs
silverjam: Do you mean benchmark the speed of upload/download, or how many resources are used by the upload/download process?

src/esthri/src/main.rs
silverjam: there's no special format for these parameters, but updated the CLI docs to reflect that

src/esthri/src/ops/download.rs
silverjam: Wrapped the new blocking stuff in spawn_blocking

silverjam: Wrapped all the blocking writes I could find, moved std::fs/std::io references to an internal module called "bio" so it was obvious if something was interacting with blocking I/O.  Using `tokio::task::spawn_blocking` had the downside of requiring a `'static` lifetime bound, which I've learned (recently) is not the same as `&'static` but this bound ended up propagating to a few odd places (most notably: `upload_from_reader` now requires that you give it ownership of the reader, but this was mostly an internal function anyway).  Might be able to get rid of this bound if we switched to `tokio::fs` but that's probably a bigger change.

src/esthri/tests/integration/upload_test.rs
john-michaelburke: I meant the speed to more easily catch a regression of some sort. I'm not sure how heavily this will be in use but if we build infrastructure around it, a benchmark with upper/lower thresholds could make it easier to lazily keep tabs on it's functioning as intended.

notoriaga (approved):

john-michaelburke (approved):

PR #206, opened 2021-10-01
[DEVINFRA-514] Improve UI for the file browser 

samvrlewis:
Adds some basic styling to improve the UI of the file browser, using Bootstrap for styles and FontAwesome for a few icons. There's also a small usability tweak to make navigation slightly easier by adding breadcrumb type links in the path heading.

With these changes, the file browser now looks like this (the blue button on the right side does the same thing as the 'tgz' before). Open to any feedback on how to make it look better or more usable.

Screenshot from 2021-10-01 17-00-31

I also experimented with serving the static CSS files directly from the esthri binary (see https://github.com/swift-nav/esthri/commit/132d2fa434727533c99bfcaaaff5cb3f1bbec004 - now reverted) to avoid having a dependency on a third party CDN to serve the CSS styles. I've reverted it for now because I'm not sure if this is something that we care about, and because it adds a lot of bulk to the repo. If others think it is worth doing, an alternative I've just realised might work well would be to download the css assets only at build time, rather than having them in the repo.


john-michaelburke (approved):

silverjam:
> With these changes, the file browser now looks like this (the blue button on the right side does the same thing as the 'tgz' before). Open to any feedback on how to make it look better or more usable.

Looks great.  A few additional tweaks I think would be great:
- Remove the `.` and `..` entries they seem superfluous with the breadcrumbs (I wanted to remove the archive downloaded from `..` since it can potentially start archive really large directories of unknown size).
- Move the tgz icon for `.` to the top bar next to the breadcrumbs
- Add a hover tooltip for the archive buttons explaining what they do 

> I also experimented with serving the static CSS files directly from the esthri binary (see [132d2fa](https://github.com/swift-nav/esthri/commit/132d2fa434727533c99bfcaaaff5cb3f1bbec004) - now reverted) to avoid having a dependency on a third party CDN to serve the CSS styles. I've reverted it for now because I'm not sure if this is something that we care about, and because it adds a lot of bulk to the repo. If others think it is worth doing, an alternative I've just realised might work well would be to download the css assets only at build time, rather than having them in the repo.

I think using a CDN is fine, if you're reaching out the AWS, you likely have access to the internet, so I don't think this is a big deal.

samvrlewis:
Nice ideas! Have implemented now:

Screenshot from 2021-10-04 09-57-18

silverjam (approved):

PR #214, opened 2021-10-29
[DEVINFRA-586] Fix some incompatabilities with AWS tool

samvrlewis:
Fixes some edge cases with the AWS tool compatibility mode, mostly to do with how directories are handled.

Allows:

- `aws cp s3://bucket/prefix/file /mydir/` -> file is downloaded to `/mydir/file`
- `aws cp /mydir/file s3://bucket/prefix/` -> file is uploaded to `s3://bucket/prefix/file`
- `aws cp sync s3://bucket/prefix /mydir` -> files within prefix are synced to `/mydir` (if either paths are to files then this will still fail, but allows flexibility in not having the trailing slash)

These are all to allow SITL functionality to continue to function as before when using the esthri compatibility mode.

Also adds some integration tests to ensure that the CLI program works as expected in the aws compatibility mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert_cmd"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


silverjam (approved):
Lgtm

PR #217, opened 2021-11-08
File browser: Allow display of compressed files [DEVINFRA-513]

samvrlewis:
Allows the file browser to display gzip compressed files, through
serving them with the content-encoding set to gzip. The way this works
is:

- Files are detected as gzip compressed by searching for the '.gz'
  suffix on the file
- For gzip compressed files, the browser will add links to the 'non
  compressed' (sans .gz) file path with a 'from_gz_compressed' query
  param
- With the 'from_gz_compressed' query param selected, the file browser
  will get the file with gz suffix and serve it, setting the content-type to 
  gzip

Also added is the ability to automatically fall back to serving a .gz
file if the path is not available. This is useful for users of esthri
who have uploaded files that have transparently been compressed (and had
a .gz suffix added) so that they can still embed links between files.

In terms of UI, the archive buttons is now show on gzip uploaded files. Clicking on the button downloads the gzip file directly.

image

### DIFF
--- a/src/esthri/src/http_server.rs
+++ b/src/esthri/src/http_server.rs
@@ -526,7 +526,7 @@ fn format_title(bucket: &str, path: &str) -> Markup {
                         (component)
                     }
 
-                    (format_archive_download_button(".", component))
+                    (format_archive_download_button(".?archive=true",&format!("Download tgz archive of {}", component)))
                 } @else {
                     a href=("../".repeat(path_length - 1 - i)) {
                         (component)
--- a/src/esthri/src/http_server.rs
+++ b/src/esthri/src/http_server.rs
@@ -482,6 +482,9 @@ fn format_object_link(path: &str) -> Markup {
                 a href=(format!("{}?from_gz_compressed=true", uncompressed_path)) {
                     (uncompressed_path)
                 }
+                span style="color: lightgrey" {
+                    " .gz"
+                }
             } @else {
                 a href=(path) {
                     (path)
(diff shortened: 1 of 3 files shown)


silverjam:
@samvrlewis How hard would it be to add some quick tests for this?

samvrlewis:
> @samvrlewis How hard would it be to add some quick tests for this?

Have added a few now :smile:

silverjam (commented):
Tested with the following:

```
esthri sync  --source 's3://drive-test-logs-2/2021-11/2021-11-04-BayLoop/plots/CDFs/' --destination 2021-11/2021-11-04-BayLoop/plots/CDFs/
esthri sync --compress --source 2021-11/2021-11-04-BayLoop/plots/CDFs/ --destination s3://esthri-test/2021-11/2021-11-04-BayLoop/plots/CDFs/
esthri serve --bucket esthri-test 
```

Seems to work well, but noticed that the archive download icon for the folder at the top of the page doesn't seen to create link for downloading an archive, just links again to the folder:

image

Also, would be nice to add some kind of indication that we're "transforming" the real file by having full original filename displayed somehow, for example:

gz

This adds a grey ".gz" that's not clickable to the filename to indicate the "true" file name.

samvrlewis:
> Seems to work well, but noticed that the archive download icon for the folder at the top of the page doesn't seen to create link for downloading an archive, just links again to the folder:

Good catch, I'd change the API to that function but had missed one of its usages. :man_facepalming: 
 
> Also, would be nice to add some kind of indication that we're "transforming" the real file by having full original filename displayed somehow, for example:

I like this, nice idea! Have added this now too:

image

silverjam (approved):

PR #226, opened 2021-11-29
sync: Add a "compressed down" option [DEVINFRA-589]

samvrlewis:
Syncing down (from bucket to local path) with compression enabled will
now automatically decompress any (gzip) compressed files. Any files that
are not already compressed will be synced as normal.

Equality checks between remote and local files (which are needed in
order to stop remotely compressed files from being re-downloaded) are
implemented by recompressing local files before syncing. This will incur
a processing cost to save a bandwidth cost. On my laptop, compressing a
1GB random file takes about 30 seconds:

    sam@sambuntu:/tmp$ head -c 1G /dev/urandom > test.random
    sam@sambuntu:/tmp$ time gzip -k test.random

    real        0m27.098s
    user        0m26.474s
    sys         0m0.513s

For my slow internet speeds, this will be better than syncing the whole
file down each time but the equation may be different depending on the
types of files synced, and processing/bandwidth constraints.

If this method of checking for file equality becomes painful for some
use-cases, one way to improve it might be to maintain a cache of md5sums
of uncompressed files to md5sums of compressed files, perhaps in a local
`.esthri` folder. This way, if a particular directory is being regularly
synced then each uncompressed file would not need to be compressed each
time.

### DIFF
--- a/src/esthri/src/compression.rs
+++ b/src/esthri/src/compression.rs
@@ -24,6 +24,7 @@ use tokio::task;
 
 use super::{Error, Result, EXPECT_SPAWN_BLOCKING};
 
+use crate::Error::FileNotCompressed;
 /// Internal module used to call out operations that may block.
 mod bio {
     pub(super) use std::fs;
@@ -102,18 +103,16 @@ pub(crate) fn path_to_compressed_path(path: &Path) -> PathBuf {
     file_gz
 }
 
-pub(crate) fn compressed_path_to_path(path: &Path) -> PathBuf {
+pub(crate) fn compressed_path_to_path(path: &Path) -> Result<PathBuf> {
     let mut pathbuf = path.to_path_buf();
-    let ext = path.extension();
-
-    if let Some(ext) = ext {
-        if ext == "gz" {
-            let filename = path.file_name().unwrap();
-            pathbuf.set_file_name(filename.to_string_lossy().strip_suffix(".gz").unwrap());
-        }
+    let ext = path.extension().ok_or(FileNotCompressed)?;
+    let filename = path.file_name().ok_or(FileNotCompressed)?;
+    if ext == "gz" {
+        pathbuf.set_file_name(filename.to_string_lossy().strip_suffix(".gz").unwrap());
+        Ok(pathbuf)
+    } else {
+        Err(FileNotCompressed)
     }
-
-    pathbuf
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


src/esthri/src/compression.rs
```diff
@@ -89,10 +99,40 @@ pub(crate) async fn compress_and_replace(path: impl AsRef<Path>) -> Result<PathB
...
+    file_gz
+}
+
+pub(crate) fn compressed_path_to_path(path: &Path) -> PathBuf {
+    let mut pathbuf = path.to_path_buf();
+    let ext = path.extension();
+
+    if let Some(ext) = ext {
+        if ext == "gz" {
+            let filename = path.file_name().unwrap();
```
john-michaelburke, line 111: Should probably use expects here or just some nested if lets.

src/esthri/src/ops/download.rs
```diff
@@ -538,7 +539,23 @@ where
...
 
-    download(s3, bucket, &key, &dest_path).await?;
+    let file_is_compressed = dest_path.ends_with(".gz");
+
+    if decompress && file_is_compressed {
+        #[cfg(feature = "compression")]
+        {
+            let dest_path = dest_path
+                .strip_suffix(".gz")
+                .expect("Should have a gz suffix");
```
john-michaelburke, line 549: Maybe this would be more appropriate as an Err since this returns a Result.

src/esthri/src/ops/sync.rs
```diff
@@ -483,38 +534,57 @@ where
...
-                            "etag mismatch: {}, local etag={}, remote etag={}",
-                            path.display(),
-                            local_etag,
-                            metadata.e_tag
-                        );
-                        download_with_dir(&s3, &bucket, &key, &metadata.s3_suffix, &directory)
+        .map(
+            |(s3, bucket, key, directory, decompress, entry)| async move {
+                let (path, local_etag, metadata) = entry;
+                let metadata = metadata.unwrap();
```
john-michaelburke, line 544: I see that this was more or less copy paste from the original work but seeing as how it is being touched, might also want to use an Err here.

src/esthri/src/ops/download.rs
samvrlewis: There might be a nicer way to do what I'm trying to achieve here but this block will only ever be entered if the suffix is `.gz` (see the check on line 542) so this line should never error. I was thinking that having this as an expect rather than an Err would be more illustrative of this, since it isn't expected to fail.

src/esthri/src/ops/sync.rs
samvrlewis: Yeah, I think this line is only highlighted in the diff because of indentation changes from `cargo format`. I didn't write the original code but the unwrap kind of makes sense to me here because we don't expect that the metadata will ever be none in this part of the code as it's always set in `flattened_object_listing`.

I guess it might be slightly nicer if the input stream here took `ListingMetadata` not as an option to signify this but then there'd be complexities sharing other parts of the code (particularly `map_paths_to_etags` and `local_to_remote_sync_tasks`) to handle it. The upload/download paths are probably getting distinct enough that maybe it's worth doing a bit of a refactor here to make that nicer, but I'll leave that for a different PR.

src/esthri/src/ops/download.rs
john-michaelburke: That sounds reasonable to me

john-michaelburke (approved):

src/esthri/src/compression.rs
samvrlewis: Have fixed this up now to return a result, which gets rid of some of these nastier `unwraps` :slightly_smiling_face:

john-michaelburke (approved):

PR #229, opened 2021-12-01
Add metadata to files compressed by esthri [DEVINFRA-589]

samvrlewis:
Adds metadata to all files that esthri uploads in compressed mode. This
currently isn't used anywhere but may be useful in the future to do
things like:

- Distinguish between files compressed by Esthri and files independently
  compressed (which may be useful for implementing transparent
  compression/decompression)
- The type of compression used by Esthri, which perhaps may change in
  the future in a non-backwards compatible way

### DIFF
--- a/src/esthri/src/compression.rs
+++ b/src/esthri/src/compression.rs
@@ -13,6 +13,7 @@
 #![cfg_attr(feature = "aggressive_lint", deny(warnings))]
 
 use std::{
+    collections::HashMap,
     ffi::OsString,
     path::{Path, PathBuf},
 };
@@ -32,6 +33,15 @@ mod bio {
     pub(super) use std::io::Seek;
     pub(super) use std::io::SeekFrom;
     pub(super) use tempfile::NamedTempFile;
+}
+
+pub(crate) fn compressed_file_metadata() -> HashMap<String, String> {
+    let mut m = HashMap::new();
+    m.insert(
+        "esthri_compress_version".to_string(),
+        env!("CARGO_PKG_VERSION").to_string(),
+    );
+    m
 }
 
 fn rewind_file<T: bio::Seek>(file: &mut T) -> Result<()> {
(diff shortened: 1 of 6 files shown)


src/esthri/src/compression.rs
```diff
@@ -34,6 +35,15 @@ mod bio {
     pub(super) use tempfile::NamedTempFile;
 }
 
+pub(crate) fn compressed_file_metadata() -> HashMap<String, String> {
+    let mut m = HashMap::new();
```
samvrlewis, line 39: I wanted to use `once_cell` here, because this HashMap is literally the same every time, but couldn't manage to get it to work, mainly because the rustoto structs (`CreateMultipartUploadRequest` and `PutObjectRequest`) only accepting values for these metadata HashMaps despite seemingly only ever using the HashMap as a reference anyway. :man_shrugging: Maybe there's a nicer way to go about it that I'm missing?

silverjam (approved):

src/esthri/src/compression.rs
silverjam: You could use once_cell, but clone the value, but that I think that's pretty much equivalent to this.  Using an iterator to initialize the HashMap would get this on one line, e.g.:

```rust
    HashMap::<_, _>::from_iter(IntoIter::new([("esthri_compress_version".into(), env!("CARGO_PKG_VERSION").into())]))
```

Adapted from https://stackoverflow.com/questions/27582739/how-do-i-create-a-hashmap-literal

silverjam: SO answer also points to this crate: https://docs.rs/maplit/latest/maplit/

samvrlewis: Ah nice! My current naive approach looks like it might be a little more understandable/readable, so I might just stick with it for now.

silverjam: 👍

PR #231, opened 2021-12-02
download: Make directory path if it doesn't exist [DEVINFRA-590]

samvrlewis:
Allows esthri to work more like `aws s3`, which will create directories
in a download command if they don't exist.

This also changes downloads to download to a temporary file, so that the
non-existing directories only get created if the download is successful
(which mirrors the aws tool behaviour).

### DIFF
--- a/src/esthri/Cargo.toml
+++ b/src/esthri/Cargo.toml
@@ -35,6 +35,7 @@ envy = "0.4"
 async-stream = "0.3"
 serde = "1"
 serde_derive = "1"
+tempfile = "3.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


silverjam (approved):

silverjam:
Looks like maybe an update to Cargo.toml to add the tempfile library is missing

samvrlewis:
Thanks for fixing up @silverjam ! We might want to rework this a little in the future to ensure that the directory path is only created if the file download works (maybe by still saving in /tmp but then copying the tempfile over rather than using persist, which is what I think aws does), but for now I think this will be great.

PR #234, opened 2021-12-09
aws sync: Align filtering to aws tool [SITL-362]

samvrlewis:
The official `aws` tool's `sync` command implements the `--include` and
`--exclude` filters in a somewhat surprising way, which is different to
how esthri currently handles them in the following two ways:

1. In `aws`, the filters are evaluated from right to left (as they
   appear in the CLI arguments) for each file. The first matching filter
   (either exclude or include) is taken as the decision for that file.
   Currently esthri runs through excludes before includes instead.

2. In `aws`, there is always an implicit `--include *` as the last
   filter. This means `--include` statements are only useful if there is
   a `--exclude` filter. Currently esthri will only select a file if it
   matches an explicit include.

To align the `aws` compatibility mode with the real `aws` tool, this
change makes esthri follow these points by slightly changing the
interface to the `sync` function. API users of this function will now
need to provide a slice of `GlobFilter` enums, where each glob is
specified as either `Include` or `Exclude` and where the order in the
slice determines the order that the filter will be evaluated. Existing
users (just SITL, I think) should be able to retain the same functional
behavior by having this slice be `[exclude filters, include filters]`.

For the second point above, this is implemented at the CLI interface as
to not break existing users of the `sync` API, that might rely on only
explicitly included files being selected.

### DIFF
--- a/src/esthri/src/ops/sync.rs
+++ b/src/esthri/src/ops/sync.rs
@@ -32,13 +32,30 @@ use crate::{
 
 type MapEtagResult = Result<(PathBuf, Result<String>, Option<ListingMetadata>)>;
 
+#[derive(Debug, Clone)]
+pub enum GlobFilter {
+    Include(Pattern),
+    Exclude(Pattern),
+}
+
+/// Syncs between S3 prefixes and local directories
+///
+/// # Arguments
+///
+/// * `s3` - S3 client
+/// * `source` - S3 prefix or local directory to sync from
+/// * `destination` - S3 prefix or local directory to sync to
+/// * `glob_filter` - An (optional) slice of filters that specify whether files
+///                   should be included or not. These are processed in order,
+///                   with the first matching filter determining whether the
+///                   file will be included or excluded. If not supplied, then
+///                   all files will be synced.
 #[logfn(err = "ERROR")]
 pub async fn sync<T>(
     s3: &T,
     source: S3PathParam,
     destination: S3PathParam,
-    includes: Option<&[impl AsRef<str>]>,
-    excludes: Option<&[impl AsRef<str>]>,
+    glob_filters: Option<&[GlobFilter]>,
     #[cfg(feature = "compression")] compressed: bool,
 ) -> Result<()>
 where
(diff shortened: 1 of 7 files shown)


src/esthri/src/ops/sync.rs
```diff
@@ -47,38 +64,16 @@ where
...
-                Err(e) => {
-                    return Err(Error::GlobPatternError(e));
-                }
-                Ok(p) => {
-                    glob_includes.push(p);
-                }
+    let filters: Vec<GlobFilter> = match glob_filters {
+        Some(filters) => {
+            let mut filters = filters.to_vec();
+            if filters.iter().any(|x| matches!(x, GlobFilter::Include(_))) {
```
silverjam, line 70: If I'm reading this right it says "if there's any include in the list, also push an include of `*`" -- wouldn't that cause it to always include everything?

src/esthri/src/ops/sync.rs
```diff
@@ -153,25 +129,25 @@ where
...
-        for pattern in glob_includes {
-            if pattern.matches(path.to_string_lossy().as_ref()) {
-                included = true;
-                break;
+        for pattern in filters {
+            match pattern {
+                GlobFilter::Include(filter) => {
+                    if filter.matches(path.to_string_lossy().as_ref()) {
+                        included = true;
+                        break;
```
silverjam, line 142: In order to implement the "left to right" processing that aws cli docs describe, don't we need to continue to process filters instead of breaking out?

src/esthri/src/ops/sync.rs
samvrlewis: Yep, you're right - I noticed that right after I pushed as well. Was supposed to be

```suggestion
            if !filters.iter().any(|x| matches!(x, GlobFilter::Include(_))) {
```
but I missed the negation. :man_facepalming: Currently fixing and writing a test for it!

src/esthri/src/ops/sync.rs
samvrlewis: Hmm, I interpreted the docs a little differently:

> Any number of these parameters can be passed to a command. You can do this by providing an --exclude or --include argument multiple times, e.g. --include "*.txt" --include "*.png". When there are multiple filters, the rule is the filters that appear later in the command take precedence over filters that appear earlier in the command. For example, if the filter parameters passed to the command were
>
> `--exclude "*" --include "*.txt"`
>
> All files will be excluded from the command except for files ending with .txt However, if the order of the filter parameters was changed to
>
> `--include "*.txt" --exclude "*"`
>
> All files will be excluded from the command.

(from https://docs.aws.amazon.com/cli/latest/reference/s3/index.html#use-of-exclude-and-include-filters)

I read that to mean that it goes "right to left" and takes the first matching rule as the decision for the file. So, once you have a match then you can break out and say that's the decision. For the `aws s3` case, if there's no match then you fall through to the implicit `--include *` case and the file will be included.

src/esthri/src/ops/sync.rs
silverjam: 👍

src/esthri/src/ops/sync.rs
silverjam: Your interpretation makes sense, but in that case, shouldn't we reverse the list of filters prior to processing them? (Or are we already)?

samvrlewis: Yeah, it's not super obvious but the list of filters is sorted in order from last specified to first when it's read from the CLI:

https://github.com/swift-nav/esthri/blob/c68a917bf8b33a2cf9646bb6396c716b5da5f91a/src/esthri/src/main.rs#L340-L375

IMO having the API accept the filters in the order that they'll be processed makes more sense than having it process the list in reverse order, so I reversed the order there.

silverjam: Ok, I was struggling to figure out where the reverse was happening, finally found it here: `filters.sort_by(|a, b| b.0.cmp(&a.0));` 😅

Can we try `filters.sort_by(|a, b| a.0.cmp(&b.0).reverse());` instead?

silverjam: Docs: https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.reverse

samvrlewis: Good call! done

silverjam (approved):

PR #238, opened 2021-12-17
Transparent compression using metadata [DEVINFRA-594] [DEVINFRA-595] [DEVINFRA-596] [DEVINFRA-597]

samvrlewis:
# Design Notes
This is a fairly large change to how esthri handles compression. Previously, esthri would append a suffix of `.gz` to all compressed files on upload, then strip this suffix when decompressing files on download. The main weakness of this is that it was hard to do this in a way that was transparent to the user. If a user were to run `aws cp myfile s3://mybucket/` with compression on then this file would exist on S3 at key `myfile.gz` which meant that when they later ran `aws cp s3://mybucket/myfile` the file wouldn't exist. Changing the filename by appending the `.gz` suffix can also break the use of `--include` and `--exclude` filters.

This changes moves to _not_ changing the filename for compressed files and instead relying on S3 metadata that is added for compressed files in order to tell if they're compressed or not. This method allows users to run commands to download/upload/sync files without needing to be concerned with how the files are represented on S3. The ultimate goal is for users to be able to use esthri without ever knowing (or, really, having to worry about) whether their files are compressed or uncompressed remotely.

As part of this, I've changed some of the nomenclature around "compression" to be "transparent compression" mainly to signify that running with transparent compression in the remote to local direction will transparently decompress files as necessary. Running in the local to remote direction will still compress everything (as before), though it won't replace the source files on local disk.

There's a bunch of subtle changes to how esthri will function after this change:

- Syncing up with compression no longer replaces local files with the compressed versions.
- Compression is no longer behind a feature flag (mainly to let the code be a little more readable.)
- Running `cp` or `sync` with `--compress` (which will be later changed to `--transparent-compression`) to download files should _always_ work to get a decompressed files locally, regardless of whether the files are compressed in S3 or not. esthri will read the metadata to figure out whether it should decompress the file for the user.
- Only files compressed by esthri will be downloaded decompressed (previously any `.gz` file could be decompressed)
- From the HTTP server, compressed files are automatically served with `content-encoding: gzip`. There's no way to access them without this header.
- Download an archive from the HTTP server will add a `.gz` suffix to files within the archive that are compressed. This is to make this functionality more user friendly with the thinking that users wouldn't be relying on the filenames here matching exactly as this is only used for manual downloads - it isn't used in any automation/scripts.

The downside to this change is that a user who is downloading directly from S3 without using esthri (eg with the native `aws` tool or through the S3 console) might be confused as to why their file doesn't match with the file they uploaded. Before rolling this change out we should communicate to users that this is happening and encourage users to use `esthri` for interacting with all SITL data. 

## Todo
Some tasks to finish before this can be merged (but figured I'd open now as a WIP in case there was any feedback):

- [x] Figure out how the HTTP server will work for this change (the compression part is currently broken)
    - To keep the existing implementation, we'd need to run a HEAD on every file being listed in order to figure out if the file is compressed or not. I don't think this is ideal as it will make pages load slower.
    - I think the simplest way of doing this is to remove the ability to download compressed files as "archives" (whic really is just allowing them to be download directly without setting the `content-encoding` header) and instead always serve them with `content-encoding: gzip` if they are compressed.
        - This should still hopefully be fairly transparent to the user, as the browser should do the decompressing. It looks like other common tools like `curl` should be able to support this too (in `curl`s case, you just need to run with `--compressed` to get the uncompressed file down)
- [x] Tidy up the CLI interface
    - I'm thinking to remove the 'esthri' CLI here and to stick with only the AWS compatible CLI, adding in a `--transparent-compression` flag. Could make this smarter about falling back to the 'real' aws tool by only falling back if argv[0] != "aws" or something (though this probably makes sense to do as a separate PR).
    - Also want to add an environment variable that will cause transparent compression to be always on.
    - Also want to add a 'pull in emergency' environment variable switch that gets esthri to always fall back to aws. To aid with a smooth rollout of this feature, I think I will do this in the opposite way first (as in, always fallback unless an env var is specified) so that we can gradually try enabling this in SITL workflows one by one.
- [x] Add some more tests
- [ ] Bump to v7 (once merged)

### DIFF
--- a/src/esthri/src/http_server.rs
+++ b/src/esthri/src/http_server.rs
@@ -12,6 +12,7 @@
 
 #![cfg_attr(feature = "aggressive_lint", deny(warnings))]
 
+use std::borrow::Cow;
 use std::cell::Cell;
 use std::convert::Infallible;
 use std::net::SocketAddr;
@@ -349,14 +350,14 @@ async fn stream_object_to_archive<T: S3 + Send + Clone>(
     // hopefully make dealing with (transparently) compressed files less
     // confusing
     let path = if obj_info.is_esthri_compressed() {
-        format!("{}.gz", path)
+        Cow::Owned(format!("{}.gz", path))
     } else {
-        path.to_string()
+        Cow::Borrowed(path)
     };
 
     let mut stream_reader = stream.into_async_read().compat();
     if let Err(err) = archive
-        .append_data(&mut header, path, &mut stream_reader)
+        .append_data(&mut header, &*path, &mut stream_reader)
         .await
     {
         abort_with_error(
--- a/src/esthri/src/http_server.rs
+++ b/src/esthri/src/http_server.rs
@@ -344,6 +344,16 @@ async fn stream_object_to_archive<T: S3 + Send + Clone>(
     header.set_mode(ARCHIVE_ENTRY_MODE);
     header.set_mtime(obj_info.last_modified.timestamp() as u64);
     header.set_size(obj_info.size as u64);
+
+    // Appending the suffix to the path within the archive should
+    // hopefully make dealing with (transparently) compressed files less
+    // confusing
+    let path = if obj_info.is_esthri_compressed() {
+        format!("{}.gz", path)
+    } else {
+        path.to_string()
+    };
+
     let mut stream_reader = stream.into_async_read().compat();
     if let Err(err) = archive
         .append_data(&mut header, path, &mut stream_reader)
--- a/src/esthri/src/http_server.rs
+++ b/src/esthri/src/http_server.rs
@@ -75,14 +75,12 @@ use serde_derive::{Deserialize, Serialize};
 
 const ARCHIVE_ENTRY_MODE: u32 = 0o0644;
 const LAST_MODIFIED_TIME_FMT: &str = "%a, %d %b %Y %H:%M:%S GMT";
-const COMPRESSION_SUFFIX: &str = ".gz";
 
 #[derive(Deserialize)]
 struct DownloadParams {
     archive: Option<bool>,
     archive_name: Option<String>,
     prefixes: Option<S3PrefixList>,
-    from_gz_compressed: Option<bool>,
 }
 
 #[derive(Debug)]
@@ -469,30 +467,12 @@ fn format_prefix_link(prefix: &str) -> Markup {
 
 #[allow(clippy::branches_sharing_code)]
 fn format_object_link(path: &str) -> Markup {
-    let is_gzip_compressed = path.ends_with(COMPRESSION_SUFFIX);
-    let uncompressed_path = match is_gzip_compressed {
-        true => path.strip_suffix(COMPRESSION_SUFFIX).unwrap(),
-        false => path,
-    };
-
     html! {
         div {
             i class="far fa-file fa-fw pe-4" { }
-            @if is_gzip_compressed {
-                a href=(format!("{}?from_gz_compressed=true", uncompressed_path)) {
-                    (uncompressed_path)
-                }
-                span style="color: lightgrey" {
-                    ".gz"
-                }
-            } @else {
                 a href=(path) {
                     (path)
                 }
-            }
-        }
-        @if is_gzip_compressed {
-            (format_archive_download_button(path, &format!("Download {}" ,path)))
         }
     }
 }
@@ -636,26 +616,10 @@ async fn item_pre_response<'a, T: S3 + Send>(
     s3: &T,
     bucket: String,
     path: String,
-    serve_compressed: bool,
     if_none_match: Option<String>,
     mut resp_builder: response::Builder,
 ) -> Result<(response::Builder, Option<(String, String)>), warp::Rejection> {
-    let (obj_info, path, serve_compressed) = match get_obj_info(s3, &bucket, &path).await {
-        Ok(info) => Ok((info, path, serve_compressed)),
-        Err(error) => {
-            if error.is_not_found() && !path.ends_with(COMPRESSION_SUFFIX) {
-                // Allows automatic fallback to the compressed version, if the
-                // non compressed version is requested. This is useful for users
-                // who might be using esthri in 'transparent compression mode',
-                // where they want to embed links between files and are unaware
-                // the files they have uploaded have been compressed.
-                let gz_path = format!("{}{}", &path, COMPRESSION_SUFFIX);
-                Ok((get_obj_info(s3, &bucket, &gz_path).await?, gz_path, true))
-            } else {
-                Err(error)
-            }
-        }
-    }?;
+    let obj_info = get_obj_info(s3, &bucket, &path).await?;
 
     let not_modified = if_none_match
         .map(|etag| etag.strip_prefix("W/").map(String::from).unwrap_or(etag))
@@ -667,7 +631,7 @@ async fn item_pre_response<'a, T: S3 + Send>(
         Ok((resp_builder, None))
     } else {
         resp_builder = resp_builder.header(CONTENT_LENGTH, obj_info.size);
-        resp_builder = resp_builder.header(ETAG, obj_info.e_tag);
+        resp_builder = resp_builder.header(ETAG, &obj_info.e_tag);
         resp_builder = resp_builder.header(CACHE_CONTROL, "private,max-age=0");
         resp_builder = resp_builder.header(
             LAST_MODIFIED,
@@ -677,17 +641,11 @@ async fn item_pre_response<'a, T: S3 + Send>(
                 .to_string(),
         );
 
-        if serve_compressed {
+        if obj_info.is_esthri_compressed() {
             resp_builder = resp_builder.header(CONTENT_ENCODING, "gzip");
         }
 
-        let content_path = if serve_compressed {
-            path.strip_suffix(COMPRESSION_SUFFIX).unwrap_or(&path)
-        } else {
-            &path
-        };
-
-        let mime = mime_guess::from_path(content_path);
+        let mime = mime_guess::from_path(&path);
         let mime = mime.first();
         if let Some(mime) = mime {
             resp_builder = resp_builder.header(CONTENT_TYPE, mime.essence_str());
@@ -808,23 +766,8 @@ async fn download(
             Err(e) => (Some(Body::from(e.to_string())), header),
         }
     } else {
-        let serve_compressed = params.from_gz_compressed.unwrap_or(false);
-
-        let path = if serve_compressed {
-            format!("{}{}", path, COMPRESSION_SUFFIX)
-        } else {
-            path
-        };
-
-        let (resp_builder, create_stream) = item_pre_response(
-            &s3,
-            bucket,
-            path,
-            serve_compressed,
-            if_none_match,
-            resp_builder,
-        )
-        .await?;
+        let (resp_builder, create_stream) =
+            item_pre_response(&s3, bucket, path, if_none_match, resp_builder).await?;
         if let Some((bucket, path)) = create_stream {
             let stream = create_item_stream(s3.clone(), bucket, path).await;
             (Some(Body::wrap_stream(stream)), resp_builder)
--- a/src/esthri/src/main.rs
+++ b/src/esthri/src/main.rs
@@ -119,7 +119,6 @@ enum EsthriCommand {
     /// Upload an object to S3
     Put {
         /// Should the file be compressed during upload
-
         #[structopt(long)]
         compress: bool,
         /// The target bucket (example: my-bucket)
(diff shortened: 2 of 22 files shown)


src/esthri/src/ops/sync.rs
```diff
@@ -464,12 +488,19 @@ where
                     for entry in entries {
                         let entry = entry.unwrap_object();
                         debug!("key={}", entry.key);
+
+                        let compressed = if transparent_decompress {
+                            head_object_request(s3, bucket, &entry.key).await?.expect("No head info?").is_esthri_compressed()
```
notoriaga, line 493: Not sure how much overhead this head object request is adding, but I think there is an opportunity for a bit more concurrency here if we need it. Right now we're making one head object request at a time but could do them at the same time, something like

```rust
fn flattened_object_listing<'a, ClientT>(
    s3: &'a ClientT,
    bucket: &'a str,
    key: &'a str,
    directory: &'a Path,
    filters: &'a [GlobFilter],
    transparent_decompress: bool,
) -> impl Stream<Item = Result<(String, Option<ListingMetadata>)>> + 'a
where
    ClientT: S3 + Send + Clone,
{
    async_stream::stream! {
    let mut stream = list_objects_stream(s3, bucket, key);
        loop {
            let entries = match stream.try_next().await {
                Ok(Some(entries)) => entries,
                Ok(None) => break,
                Err(err) => {
                    yield Err(err);
                    return;
                }
            };
            let mut entries = futures::stream::iter(entries)
                .map(|entry| async move {
                    let entry = entry.unwrap_object();
                    debug!("key={}", entry.key);
                    let compressed = if transparent_decompress {
                        head_object_request(s3, bucket, &entry.key)
                            .await?
                            .expect("No head info?")
                            .is_esthri_compressed()
                    } else {
                        false
                    };
                    Ok((entry, compressed))
                })
                .buffered(5);
            loop {
                let (entry, compressed) = match entries.try_next().await {
                    Ok(Some(entry)) => entry,
                    Ok(None) => break,
                    Err(err) => {
                        yield Err(err);
                        return;
                    }
                };
                let path_result = Path::new(&entry.key).strip_prefix(key);
                if let Ok(s3_suffix) = path_result {
                    if process_globs(&s3_suffix, filters).is_some() {
                        let local_path: String = directory.join(&s3_suffix).to_string_lossy().into();
                        let s3_suffix = s3_suffix.to_string_lossy().into();
                        yield Ok((
                            local_path,
                            ListingMetadata::some(s3_suffix, entry.e_tag, compressed),
                        ));
                    }
                } else {
                    yield Err(path_result.err().unwrap().into());
                    return;
                }
            }
        }
    }
}
```
Not 100% sure that will work but the main point is to process all the head requests for each batch of entries at the same time

notoriaga (commented):
Looks great so far! I wonder how much impact this will have on bandwidth used in the SITL cluster. iirc we were bumping into some limits with that in the past

src/esthri/src/http_server.rs
```diff
@@ -346,6 +344,16 @@ async fn stream_object_to_archive<T: S3 + Send + Clone>(
...
     header.set_size(obj_info.size as u64);
+
+    // Appending the suffix to the path within the archive should
+    // hopefully make dealing with (transparently) compressed files less
+    // confusing
+    let path = if obj_info.is_esthri_compressed() {
+        format!("{}.gz", path)
+    } else {
+        path.to_string()
+    };
```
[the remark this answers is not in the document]
samvrlewis: > So the situation here is someone is requesting an archive that itself might contain transparently compressed files? Adding the extension makes sense otherwise yeah I could see people wondering why they can't open the file etc.

Yep, exactly!

> What you have here will actually allocated an owned value in both cases. The first one (adding the .gz) is unavoidable. For the second one though, the string is fine as it, but because the first branch returns a string you need to return in the second branch as well. An alternative would be something like

I did actually spend some time mulling over this bit of code as I didn't like having to do the allocation in `path.to_string()`. I was thinking of doing something like

```
    let tmp_string; 
    let path = if obj_info.is_esthri_compressed() {
        tmp_string = format!("{}.gz", path);
        &tmp_string
    } else {
        path
    };
```

Which I _think_ is roughly equivalent to the `Cow` approach (but probably uglier), but ultimately decided against it because I thought the performance penalty wasn't worth the uglification of the code :smiley: . 

The `Cow` pattern is cool though, thanks - seems like the right idiomatic (rustonic?) way of doing things. I'll update to use it. I have to say though, having the type named `Cow` is a bit confusing for my small brain when it's used for this purpose - if I saw `Cow`s in the code I'd immediately think that the code was trying to do clone on write optimization and then get confused when there's no mutation going on. I guess I'll know better from now on!

src/esthri/src/main.rs
```diff
@@ -501,6 +495,12 @@ async fn async_main() -> Result<()> {
     env_logger::init();
 
     let aws_compat_mode = is_aws_compatibility_mode();
+    let always_fallback = should_always_fallback_to_aws();
+
+    if always_fallback && aws_compat_mode {
```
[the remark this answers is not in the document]
samvrlewis: I guess you could probably argue either way makes sense, but I was thinking here that if you're not in the AWS compatibility mode then falling back to the real AWS tool all of the time probably makes less sense, as you might want to use the `EsthriCli`

This logic is getting a bit confusing here though, I think it's probably ripe for a refactor. I'd like to unify the AWS CLI and the esthri CLI into one CLI that's basically the `aws s3` commands we support and then some other commands (like `abort-upload`). At the moment we have multiple ways of doing the same thing from the CLI (for eg: `get` in the `EsthriCli` and `cp` in the `AwsCli`) which is probably unnecessary.

src/esthri/src/ops/sync.rs
samvrlewis: Aha, this is great, much nicer! Will definitely speed up syncs with lots of files.

samvrlewis:
Hacked this into a SITL PR and deployed to staging (https://github.com/swift-nav/sitl/pull/588)

On the `starbuck_icbins_workflow` it looks like it worked well: https://sitl-workflows-staging.ce.swiftnav.com/run-results/1076#executions can't even really tell that the files are compressed, unless you download an archive of all files.

Will do more tests against some other popular workflows but looks promising so far. :crossed_fingers:

samvrlewis:
Think this is good to be be merged once approved - the SSRW run I ran against staging (https://sitl-workflows-staging.ce.swiftnav.com/run-results/1082#executions) also looked good (workflow passed, file browser worked to transparently serve files, `sitl --staging runs sync-results 1082` synced down the files uncompressed) but feel free to see if you can find anything wrong with it.

src/esthri/src/http_server.rs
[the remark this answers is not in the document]
notoriaga: > if I saw `Cow`s in the code I'd immediately think that the code was trying to do clone on write optimization and then get confused when there's no mutation going on

Yeah it's a little confusing because that is what `Cow` _is_ for, but it just so happens that it also solves a pretty prevalent problem in rust where you want to return some data from a scope (typically a function but in your case just a block) that is either owned or borrowed. This brief blog does a good job explaining the different usecases - https://deterministic.space/secret-life-of-cows.html

notoriaga:
> Think this is good to be be merged once approved - the SSRW run I ran against staging (https://sitl-workflows-staging.ce.swiftnav.com/run-results/1082#executions) also looked good (workflow passed, file browser worked to transparently serve files, `sitl --staging runs sync-results 1082` synced down the files uncompressed) but feel free to see if you can find anything wrong with it.

Noticed it took 5 hours to run on staging (we only run one worker of each type on staging). If you find yourself needing to do something similar in the future you can also edit the hpa's to allow more workers. For example

```
kubectl -n sitl-workflows-staging edit hpa sitl-orch-fan-out-scaler
```

Then change `spec.maxReplicas` to 10 or something. Just make sure to change it back after :smile:

src/esthri/src/main.rs
```diff
@@ -244,17 +241,22 @@ fn is_aws_compatibility_mode() -> bool {
     is_run_as_aws || has_aws_env_var
 }
 
+/// Allows for an escape hatch to be set that triggers always falling back from
+/// esthri into the real aws tool.
+fn should_always_fallback_to_aws() -> bool {
+    env::var("ESTHRI_AWS_ALWAYS_FALLBACK").is_ok()
```
notoriaga, line 247: Would it be possible to switch this to check that it's equal to `true` or `1` etc? There isn't an easy way to unset a variable from a workflow config, so workflows would have to write python/bash to do it. But if it's a true/false thing you could make it an `env` resource and toggle it. Or maybe we could make it a proper field in the workflow config?

notoriaga (approved):
What was the plan again r.e. using this with wolf? We decided just to integrate with the code in sitl correct (judging by the other pr)? I don't think this set up with the env vars will work well if we wanna enable it for wolf

src/esthri/src/main.rs
notoriaga: If we go with adding it to the config proper than unsetting it would be fine

src/esthri/src/main.rs
```diff
@@ -385,16 +384,10 @@ async fn dispatch_esthri_cli(cmd: EsthriCommand, s3: &S3Client) -> Result<()> {
...
-            }
-            #[cfg(not(feature = "compression"))]
-            {
+
+            if matches!(cmd, Put { compress: true, .. }) {
+                upload_compressed(s3, bucket, key, file).await?;
+            } else {
                 upload(s3, bucket, key, file).await?;
             }
         }
```
notoriaga, line 393: One last tiny nit but I think you can skip these `matches!` and it's a bit clearer

```rust
match cmd {
        Put {
            ref bucket,
            ref key,
            ref file,
            compress,
        } => {
            setup_upload_termination_handler();
            if compress {
                upload_compressed(s3, bucket, key, file).await?;
            } else {
                upload(s3, bucket, key, file).await?;
            }
        },
        ...
}
```

samvrlewis:
> What was the plan again r.e. using this with wolf? We decided just to integrate with the code in sitl correct (judging by the other pr)? I don't think this set up with the env vars will work well if we wanna enable it for wolf

It should work with Wolf, as Wolf runs `aws s3 sync`
https://github.com/swift-nav/wolf/blob/3a322ca88253a1b4aa7251993145e79fcfd33738/src/Network/AWS/Wolf/Act.hs#L28

and in SITL `aws` is hardlinked to esthri https://github.com/swift-nav/sitl/blob/c1566b352e6bb3afa619517961f7df67f90c5668/docker/Dockerfile.base#L14

So as long as the `ESTHRI_AWS_COMPAT_MODE_COMPRESSION` and `ESTHRI_AWS_COMPAT_MODE` env vars are set and `ESTHRI_AWS_ALWAYS_FALLBACK` is not set, then it should use transparent compression.

You can see that it worked in my staging runs (https://sitl-workflows-staging.ce.swiftnav.com/run-results/1082#executions) as all the files uploaded by Wolf in the workflows are compressed (can see by checking the response headers and seeing `content-encoding: gzip`, eg:

Screenshot from 2022-01-11 11-02-16

Or did you mean something else by 'using this with wolf' ?

src/esthri/src/main.rs
samvrlewis: That's a great point, I like the idea of using an env resource for this. I avoided it initially because I thought it'd be hard to know if the value should be `true` or `True` or `1` or `TRUE` but I guess I can probably just lowercase it in the code and check for equality to `true` or `1`. Will make it much simpler to update workflows to turn it on/off which is great.

IMO it makes a bit more sense as a env resource than as a permanent config option as it should hopefully be only a temporary thing that we have while rolling it out. Eventually hopefully all workflows can use it and don't need to turn it off.

notoriaga:
> Or did you mean something else by 'using this with wolf' ?

Sorry yeah I think the problem is that wolf runs in a different process than the worker. So when a user wants to enable compression (or disable once it's the default), I'm not quite sure how we can make that apply to wolf's s3 calls (to be clear I'm talking about `store/input`/`store/output`). 

When the worker runs and sees that it should enable/disable compression, it has the ability to change the env variables of the workflow step that it is about to start. But, I don't think we can change the env variables of wolf's process on the fly. Wolf's env is determined by a k8s deployment which I think we could edit on the fly but would be a mess

Does that make sense? Maybe I'm just thinking about this wrong

samvrlewis:
> Sorry yeah I the problem is around the enabling/disabling not it working generally. I think the problem is that wolf runs in a different process than the worker. So when a user wants to enable compression (or disable once it's the default), I'm not quite sure how we can make that apply to wolf's s3 calls (to be clear I'm talking about `store/input`/`store/output`).

Talked this over with @notoriaga on Slack and came to the agreement on a way forward:

- Get the SITL worker to upload/download from `store/input`/`store/output` rather than Wolf
  - Requires adding an option to tell wolf not to sync these directories
- Add a `compression` config option to the SITL workflow spec
  - If set, this will get the worker to sync up/down with compression and will turn on the esthri compression env variable

silverjam (approved):

PR #249, opened 2022-01-25
Memory and bandwidth improvements [DEVINFRA-609] [DEVINFRA-616] [DEVINFRA-617] [DEVINFRA-618]

notoriaga:
Pretty messy but I think there are some parts that would be easy to factor out into smaller changes. 

The biggest improvement is with get requests, the performance is roughly the same but the memory usage is way less

these changes
esthri-new-get

esthri 7.0.0
esthri-get-old

and aws cli 
aws-put
 
And calling multiple downloads at a time (i.e. sync down) scales nicely (this was 3 files)
sync-down

Uploads seem to be pretty similar, but I think could be made faster. The bufferpool stuff is probably not optimal, I think I clear/resize it too often  

This also breaks the project up into 3 crates one for the lib, one for the cli and a shared crate for tests. I mostly did this because rust-analyzer was bugging out on the project and I thought simplifying the feature stuff might help? Not sure if that was it but rust-analyzer works on it (for me) now. Also I kind of prefer multiple crates but :shrug:. This part of the refactor still needs some clean up


notoriaga:
This also removes all the `bio` stuff and tries to use non blocking apis everywhere. The two places this was tricky was with `tempfile` and `write_all_at`. For tempfile I tried to create an async wrapper, which turned out okay I think. `write_all_at` operates on a `&std::file::File` so I pass `Arc<std::file::Files>`s to `spawn_blocking` calls.

crates/esthri/src/ops/download.rs
```diff
@@ -0,0 +1,234 @@
...
+        } else {
+            path.pop();
+            Ok(path)
+        }
+    })
+    .await?
+}
+
+#[cfg(unix)]
+async fn write_all_at(file: Arc<std::fs::File>, buf: Bytes, offset: u64) -> Result<()> {
```
notoriaga, line 204: It would be interesting to add support for https://github.com/tokio-rs/tokio-uring as a feature. Should give a big boost to the write performance here

crates/esthri/src/ops/upload.rs
```diff
@@ -0,0 +1,435 @@
...
+use crate::{
+    bufferpool::{Buffer, Pool},
+    compression::compress_to_tempfile,
+    config::Config,
+    errors::{Error, Result},
+    handle_dispatch_error,
+    rusoto::*,
+};
+
+static PENDING_UPLOADS: Mutex<Vec<PendingUpload>> = parking_lot::const_mutex(Vec::new());
```
notoriaga, line 34: If I'm understanding the current impl we only store one pending multipart upload and we could have multiple during a sync. This should cancel all of them

crates/esthri/src/bufferpool.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::Config;
+
+static POOL: OnceCell<Pool> = OnceCell::new();
+
+pub struct Pool {
+    shared: Arc<Shared>,
+}
+
+impl Pool {
+    pub fn global() -> &'static Pool {
```
samvrlewis, line 16: Isn't the disadvantage of pooling like this is that once we've allocated a pool in the buffer, esthri will hold that memory until it closes? I guess this is probably OK for the CLI but as a library it seems like it could be problematic because an application might upload (causing the esthri to grab up to `concurrent_upload_tasks*upload_part_size` memory) and then do other non-esthri things afterwards. And when it's doing the non-esthri things afterwards, the OS won't be able to claim back that memory that esthri has reserved? In practice, I guess this is probably OK for the SITL worker at the moment, as it does all its uploading at the end and then exits, but it seems like it might not be the best practice (maybe we'd want to daemonify the worker at some stage which wouldn't play well with this, for eg).

Maybe I'm missing something but what benefits does having our own pool give? Quicker allocations? AFAICT, the memory use of esthri is already kind of bounded by `concurrent_upload_tasks` and `upload_part_size` so I'm not sure that this pool gives us much benefit?

notoriaga: Oh yeah we definitely should not keep this, on top of what you said this is just broken. Once I heard you had success with the uploading I just gave up on it. The downloading I have here doesn't explicitly allocate anything so doesn't use the pool

notoriaga: I think the main things worth taking from this are 

- the implementation of download impl
- the PendingUploads tracking/ctrl-c handling 
- replacement of a bunch of blocking api calls

Everything else is less of an improvement and just style kind of stuff/minor things

.gitattributes
```diff
@@ -1 +1 @@
-src/esthri/tests/data/** filter=lfs diff=lfs merge=lfs -text
```
notoriaga, line 1: I think I might of fudged up the lfs stuff by not using `git mv` to move these files

samvrlewis (approved):
Ok to merge @notoriaga?

notoriaga:
:shipit:

Maybe we could work on the etag/compression stuff in our donut?

PR #253, opened 2022-01-28
Reduce upload memory usage [DEVINFRA-609]

samvrlewis:
To have parity with the `aws` tool we currently upload files in 8MiB
chunks, which allows syncs to work consistently between both tools.

However, currently, this causes a large amount of memory use as reads
are read into 8MiB buffers and with a high amount of upload tasks, this
causes large amounts of data to sit waiting in memory to be uploaded.

This change moves to giving a stream that the rustoto upload function
consumes, that means only 1MiB should be read at a time. This reduces
the amount of memory use, especially when esthri is configured to use a
large amount of upload tasks.

--

Hopefully this shouldn't be too hard to merge into @notoriaga's other branch :grinning:

### DIFF
--- a/src/esthri/src/ops/upload.rs
+++ b/src/esthri/src/ops/upload.rs
@@ -13,14 +13,21 @@
 #![cfg_attr(feature = "aggressive_lint", deny(warnings))]
 
 use std::borrow::Cow;
-use std::sync::Mutex;
+use std::convert::TryInto;
+
+use std::io::SeekFrom;
+
+use std::sync::{Arc, Mutex};
 use std::{collections::HashMap, path::Path};
+
+use bytes::{Bytes, BytesMut};
 
 use futures::{stream, Future, Stream, StreamExt, TryStreamExt};
 
 use log::{debug, info, warn};
 use log_derive::logfn;
 use once_cell::sync::Lazy;
+
 use tokio::task;
 
 use crate::{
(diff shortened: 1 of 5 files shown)


src/esthri/src/ops/upload.rs
```diff
@@ -186,100 +193,149 @@ where
...
-                    Ok((buf, read_size, reader))
-                })
-                .await
-                .expect(EXPECT_SPAWN_BLOCKING);
+                let read_size = u64::min(remaining, read_size);
+
+                let mut buf = BytesMut::with_capacity(read_size as usize);
+                buf.resize(read_size as usize, 0);
+
+                let reader = reader.unwrap();
```
silverjam, line 229: Any danger of a panic here?

src/esthri/src/ops/upload.rs
```diff
@@ -186,100 +193,149 @@ where
...
         },
-    ))
+    ));
+
+    (to_read, stream)
 }
 
-async fn create_chunk_upload_stream<StreamT, ClientT>(
-    source_stream: StreamT,
+fn get_number_of_upload_chunks_for_file_size(file_size: u64) -> u64 {
```
silverjam, line 268: How about `chunk_count`? 🙂

samvrlewis: Hah, yes, a much more sensible name!

src/esthri/src/ops/upload.rs
samvrlewis: No - this is heavily inspired/copied from your original code - `reader` is only `None` in the stop condition, so it'll always be `Some` here.

silverjam (approved):
lgtm

@notoriaga how should we handle merging this with your refactor?

notoriaga:
I think we could go about it two ways

1. We pull in download changes + blocking call removal stuff into this pr

2. If people like the direction my branch is going in (splitting up the crates etc) it'd be better to merge this into my branch and go from there 

2 is probably slightly more work/touches more stuff etc, but both will probably be kind of awkward merges. The bio removal stuff is probably the most invasive change but also I think it's really valuable, a misplaced blocking call can lead to dead/livelock

Either way I could throw one of the two together tomorrow morning before going into the lab

notoriaga:
And I'm using the term "merge" pretty loosely here - I think 1. would be easier to do via copy/paste

samvrlewis:
Spent a little bit of time testing this out on a SITL pod (which has much more bandwidth than I do :sweat_smile:) with this [little script](https://gist.github.com/samvrlewis/6e6a9c9f9b1674c9486c680c9ae50705) that recorded RSS memory usage and bytes sent over time, to get a feel for how these changes made esthri perform relative to the previous version and the aws tool. Not a perfect benchmark (and these were only run once, so they probably shouldn't be read into too much) but they're at least a little informative.

The tldr is that everything seems OK with this change, but have included the data below as it is interesting:

## Esthri Old
esthri_old

## Esthri New
esthri_new

## aws
aws

Esthri is much faster than aws, but uses more memory. It also negates some of that speed by seemingly taking a bit longer to finalize after everything seems to be updated. Not sure if this is when it's finalizing the multipart upload or if something else is going on - might be worth investigating this further at some stage, but at least it's not a regression.

I experimented with setting the task config to get a similar memory usage to `aws`, and luckily esthri is still on par with aws.

esthri_new_tweaked

Also ran new and old under heaptrack with the default configuration as a sanity check, but these seemed to get a similar profile to my script:

## Old
Screenshot from 2022-01-28 16-40-31

## New
Screenshot from 2022-01-28 16-40-45

samvrlewis:
> Either way I could throw one of the two together tomorrow morning before going into the lab (obviously free free to go for it @samvrlewis, I just feel back about the state my pr is in and don't want you to have to deal with the mess)

Happy for you to pull this into your branch, if you're happy to do it (I guess without the buffer pool stuff?). Otherwise I'm happy to have a look at this on Monday.

I haven't looked at download memory usage at all yet either, so keen to get your changes in if that helps there as, from what I saw, we were getting OOM errors in some workers during the 'sync down' phase.

notoriaga:
> Not sure if this is when it's finalizing the multipart upload or if something else is going on - might be worth investigating this further at some stage, but at least it's not a regression.

The finalization should just be an api call right? Fees like something is probably going on

notoriaga:
And yeah the bufferpool was only for the uploads so we'll just use what you have. And from my (brief) experimenting I think we should be set with sync down, the new downloads seem to be using less than the aws cli

src/esthri/src/ops/upload.rs
```diff
@@ -186,100 +193,149 @@ where
...
+
+                let reader = reader.unwrap();
+
+                let result: Result<Bytes> = {
+                    let reader = reader.clone();
+                    let mut reader = reader.lock().unwrap();
+                    let reader = &mut *reader;
+
+                    let seek = chunk_size * chunk_number + part_number * read_size;
+                    reader.seek(SeekFrom::Start(seek)).ok()?;
```
notoriaga, line 237: `.ok()?` will convert `Err` to `None` and then `propagate` the None. This will stop the stream but the error will be lost. wrapping it up in a block with the `Result<Bytes>` thing won't work, that would need to be a try block - https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html (which is unstable)

If you wanna use `?` with errors in this situation you should use [stream::try_unfold](https://docs.rs/futures/latest/futures/stream/fn.try_unfold.html). Many stream/iterator functions/combinators  have a `try_` variant for when you have to perform fallible operations

samvrlewis:
Closing, this was merged into https://github.com/swift-nav/esthri/pull/249 in https://github.com/swift-nav/esthri/pull/254

PR #254, opened 2022-01-31
Reduce upload memory usage [DEVINFRA-609]

samvrlewis:
To have parity with the aws tool we currently upload files in 8MiB
chunks, which allows syncs to work consistently between both tools.

However, currently, this causes a large amount of memory use as reads
are read into 8MiB buffers and with a high amount of upload tasks, this
causes large amounts of data to sit waiting in memory to be uploaded.

This change moves to giving a stream that the rustoto upload function
consumes, that means only 1MiB should be read at a time. This reduces
the amount of memory use, especially when esthri is configured to use a
large amount of upload tasks.

---

This is https://github.com/swift-nav/esthri/pull/253 rebased onto https://github.com/swift-nav/esthri/pull/249


crates/esthri/src/tempfile.rs
```diff
@@ -52,6 +52,10 @@ impl TempFile {
         self.file.as_mut().unwrap()
     }
 
+    pub fn get_file(&mut self) -> File {
```
samvrlewis, line 55: I'm pretty sure this is wrong, but I'm not sure how to make it work with the `'static` lifetime needed for the upload. I think getting the File like this could run into issues as I think the temp file it's pointed to could get deleted from under it? Need to look into it further, just did the minimum to get it compiling but would welcome any ideas.

notoriaga: Hmm I believe you're pretty much spot on here. I think what you have will work fine, fwiw. In order to call `take_std_file` or `take_file` (I just renamed `get_file` -> `take_file` to be consistent with the other method that takes the file), you just have to make sure that the rest of the `TempFile` you took the file from stays alive as long as you are using the file, so for example where we use it - 

```rust
let (mut tmp, size) = compress_to_tempfile(path).await?;
upload_from_reader(
    s3,
    bucket,
    key,
    tmp.get_file(),
    size,
    Some(crate::compression::compressed_file_metadata()),
)
.await?;
Ok(())
```

`tmp` should be alive for the whole duration of `upload_from_reader`, and after that we don't need it so were good.

I agree that it's not super clean, but I don't think there is a way around it due to the static constraint you mention. In order to support a non static reader I think we'd need `ByteStream` to look like (abbreviated slightly) 

```
pub struct ByteStream<'a> {
    inner: Pin<Box<dyn Stream<Item = Result<Bytes, io::Error>> + Send + 'a>>,
}
```

As to why it's not set up that way I'm not sure. It might be a limitation of the underlying http library, hyper, or it might have just been a design choice of the rusoto authors. I'm curious if the new aws sdk also has this constraint

crates/esthri/src/ops/upload.rs
```diff
@@ -10,20 +10,21 @@
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
 
-use std::{borrow::Cow, collections::HashMap, path::Path};
+use std::convert::TryInto;
+use std::{borrow::Cow, collections::HashMap, io::SeekFrom, path::Path, sync::Arc};
 
 use bytes::{Bytes, BytesMut};
+use futures::lock::Mutex as stdMutex;
```
notoriaga, line 17: Generally once I'm "in" the tokio ecosystem I prefer to use tokio versions of things if available, so for a lock tokio::sync::Mutex. Note you don't always need to use a async mutex when working with async code (they explain that in the tokio docs https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html), but in this case you do need to, but we hold the lock "across an await point" (i.e. when we read from the file). But for example, the `PENDING_UPLOADS` lock does not do this, so it can just be a sync mutex.

notoriaga:
Btw just leaving comments about what I'm changing for posterity, I'll just push the changes directly on this branch if you don't mind :smile:

crates/esthri/src/ops/upload.rs
```diff
@@ -336,39 +335,22 @@ where
...
-                        } else {
-                            Ok(Some((
-                                buf,
-                                (remaining - read_size, part_number + 1, Some(reader)),
-                            )))
-                        }
-                    }
-                    Err(err) => Err(err),
+                    guard.read_exact(&mut slice[..read_size as usize]).await?;
+                }
```
notoriaga, line 346: No need to clone the Arc here, as long as you drop the guard after reading from the file you can return the original reader. When you think about it this makes sense, each chunk is read serially, so we only need one reader. The reader cloning comes into play reading multiple chunks at a time

notoriaga:
merged in this stuff, picking it up off on the original branch

crates/esthri/src/tempfile.rs
samvrlewis: Makes sense! I guess the only downside is something like this would still compile but would cause issues?

```
let tmp: TemporaryFile; // temp file that already has things written to the file on disk
let file_handle = tmp.take_file().unwrap()
drop(tmp) // dropping causes the temp file to be deleted on disk
// file_handle is still valid even though the temporary file it points to has been dropped & deleted
```

..which I suppose is probably why the compiler doesn't like using references here.

It is kinda annoying that the `static` stuff means it's hard to get the nice advantages we'd otherwise get by using references pointed to a `TempPath`. :disappointed:

PR #255, opened 2022-02-02
Improvements for low bandwidth connections [DEVINFRA-616]

samvrlewis:
- Changes retry to terminate after 5 attempts (have made the number of attempts configurable too), rather than after 45 seconds
- Lowers the default tasks amounts
- Cleanup of some now unused config options

### DIFF
--- a/crates/esthri/src/ops/sync.rs
+++ b/crates/esthri/src/ops/sync.rs
@@ -29,6 +29,7 @@ use crate::{
     errors::{Error, Result},
     handle_dispatch_error, list_objects_stream,
     rusoto::*,
+    tempfile::TEMP_FILE_PREFIX,
     types::ListingMetadata,
     types::{S3ListingItem, S3PathParam},
 };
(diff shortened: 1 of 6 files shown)


silverjam (approved):

silverjam:
How does this impact performance for high bandwidth connections?

notoriaga (approved):
I'm curious how it compares to the aws cli too. Apparently they default to allowing 10 requests to aws in flight https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-concurrent-requests

With these defaults we'd be at 32 right?

crates/esthri/src/ops/sync.rs
```diff
@@ -177,6 +178,9 @@ fn create_dirent_stream<'a>(
                 yield Err(entry.err().unwrap().into());
                 return;
             };
+            if entry.file_name().to_string_lossy().contains(TEMP_FILE_PREFIX) {
```
samvrlewis, line 181: A somewhat unrelated fix, but have included it here just so we don't have branches on top of branches on top of branches. A little more info in the commit message: https://github.com/swift-nav/esthri/pull/255/commits/9cdafb7629c30edf547f2ce223b21258793b170c

samvrlewis:
Some benchmarking of this here: https://swift-nav.atlassian.net/wiki/spaces/~460004640/pages/2137130948/Esthri+Benchmarking

Testing on a pod in an AWS environment (with high speed S3 transfers) esthri now is faster and either uses less memory or close to the amount of memory of the `aws tool`. Some sample results for all the non-Swift esthri devs out there. :grin: 

image

image

image

PR #260, opened 2022-02-08
don't calculate etag on sync up if the file doesn't exist

notoriaga:
From a donut with @samvrlewis - title + another sync -> async conversion

### DIFF
--- a/crates/esthri/src/ops/sync.rs
+++ b/crates/esthri/src/ops/sync.rs
@@ -40,7 +40,7 @@ use crate::{
 struct MappedPathResult {
     file_path: Box<dyn AsRef<Path>>,
     source_path: PathBuf,
-    local_etag: Result<String>,
+    local_etag: Result<Option<String>>,
     metadata: Option<ListingMetadata>,
 }
 
@@ -208,15 +208,17 @@ fn create_dirent_stream<'a>(
 }
 
 #[derive(Debug, Clone, Copy)]
-enum SyncCompressionDirection {
-    None,
+enum SyncDirection {
     Up,
+    UpCompressed,
     Down,
+
+    DownCompressed,
 }
 
 fn translate_paths<StreamT>(
     input_stream: StreamT,
-    compressed: SyncCompressionDirection,
+    compressed: SyncDirection,
 ) -> impl Stream<Item = impl Future<Output = MapPathResult>>
 where
     StreamT: Stream<Item = Result<(String, Option<ListingMetadata>)>>,
@@ -228,18 +230,30 @@ where
         let file_path: Box<dyn AsRef<Path>> = Box::new(path);
 
         let (file_path, source_path, local_etag) = match compressed {
-            SyncCompressionDirection::None => {
-                let local_etag = compute_etag(&source_path).await;
+            SyncDirection::Up => {
+                let local_etag = if metadata.is_some() {
+                    compute_etag(&source_path).await.map(Option::Some)
+                } else {
+                    Ok(None)
+                };
                 (file_path, source_path, local_etag)
             }
-            SyncCompressionDirection::Up => {
+            SyncDirection::UpCompressed => {
                 // If we're syncing up with compression, then everything
                 // should be compressed
                 let (tmp, _) = compress_to_tempfile(&source_path).await?;
-                let local_etag = compute_etag(tmp.path()).await;
+                let local_etag = if metadata.is_some() {
+                    compute_etag(&source_path).await.map(Option::Some)
+                } else {
+                    Ok(None)
+                };
                 (tmp.into_path(), source_path, local_etag)
             }
-            SyncCompressionDirection::Down => {
+            SyncDirection::Down => {
+                let local_etag = compute_etag(&source_path).await.map(Option::Some);
+                (file_path, source_path, local_etag)
+            }
+            SyncDirection::DownCompressed => {
                 // if we're syncing down with compression, then there
                 // could be both esthri compressed files and non
                 // compressed files within the prefix. We should only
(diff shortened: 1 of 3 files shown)


crates/esthri/src/ops/sync.rs
```diff
@@ -169,77 +170,86 @@ fn process_globs<'a, P: AsRef<Path> + 'a>(path: P, filters: &[GlobFilter]) -> Op
...
+                }
+                Err(e) => {
+                    tx.send(Err(e.into())).expect("failed to send error");
+                    break;
+                }
             }
         }
-    }
+    });
+    UnboundedReceiverStream::new(rx)
```
silverjam, line 205: Why do we need to spawn a thread in order to generate an async list of paths?

notoriaga: WalkDir is built on top of [ReadDir](https://doc.rust-lang.org/stable/std/fs/struct.ReadDir.html) which is blocking, so we either need an async WalkDir (tokio has an async version of ReadDir but we'd have to implement the recursive bit) or we need to consume the iterator inside a call to spawn_blocking

notoriaga: And it might not actually spawn a thread if there is an idle blocking thread ready

silverjam: Nice

crates/esthri/src/ops/sync.rs
```diff
@@ -169,77 +170,86 @@ fn process_globs<'a, P: AsRef<Path> + 'a>(path: P, filters: &[GlobFilter]) -> Op
...
                 (file_path, source_path, local_etag)
             }
-            SyncCompressionDirection::Up => {
+            SyncCmd::UpCompressed => {
                 // If we're syncing up with compression, then everything
                 // should be compressed
                 let (tmp, _) = compress_to_tempfile(&source_path).await?;
-                let local_etag = compute_etag(tmp.path()).await;
+                let local_etag = if metadata.is_some() {
+                    compute_etag(&source_path).await.map(Option::Some)
```
samvrlewis, line 242:
```suggestion
                    compute_etag(tmp.path()).await.map(Option::Some)
```

I think we still need to be computing the etag of the compressed file here.

(Probably also reveals that we might need better `sync` test coverage - although admittedly it's tricky to test what a sync command actually has done atm)

notoriaga: Ahh good catch. Yeah we could definitely improve on the testing situation

crates/esthri/src/ops/sync.rs
samvrlewis: Just for my own understanding, does that mean that previously when we did

```
let dirent_stream = create_dirent_stream(directory, filters);
let etag_stream = translate_paths(dirent_stream, compression).buffer_unordered(task_count);
```

The whole stream of futures might've been blocked while it waited on WalkDir? And practically, maybe that was okay-ish because it probably would've never blocked for long?

crates/esthri/src/ops/sync.rs
```diff
@@ -615,6 +650,7 @@ where
                             warn!("s3 etag error: {}", err);
                         }
                     },
+                    Ok(None) => warn!("no local etag"),
```
samvrlewis, line 653: This should only happen if we have a logic error somewhere, right? Would it be more appropriate to panic (or do one of your fancy panic inducing macros? :grin: )

samvrlewis (approved):

crates/esthri/src/ops/sync.rs
notoriaga: Each iteration of the `WalkDir` loop had a potential block (probably by making a call to read_dir). I think you're right in that it's not a big problem because it's all metadata calls not actual io.

> The whole stream of futures might've been blocked while it waited on WalkDir?

It kind of depends on the runtime. If you call a blocking operation in a non-blocking thread you block the runtime that you are on. In the case of a multi-threaded runtime you might not even notice because the other threads will be unaffected. 

I don't know exactly where the blocking would happen in this case with all the nested streams/buffer_unordereds by generally it'd look like something like this -

```rust
async fn f() -> i32 {
    let num = http_call().await;
    num
}

async fn bad_f() -> i32 {
    let num = http_call().await;
    blocking_call();
    num
}

async fn g() {
    // iterator of 5 futures
    let reqs = std::iter::from_fn(|| f()).take(5);
    let reqs = futures::stream::iter(reqs);
    // await them all at the same time
    reqs.buffer_unordered(5).await;
}
```

if (assuming single threaded runtime) you swapped f and bad_f you'd run into a situation where future-1 got a response from `http_call` and started `blocking_call`. Now futures 2-5 will not be able to make progress on their `http_call`s because the thread is stuck awaiting on `blocking_call`.

The tldr is that async rust is cooperative multitasking. `.await` is how you yield to the scheduler so it can make progress on other tasks. If the gap between awaits is too long then you risk starving whatever else is going on in the background of cpu time

crates/esthri/src/ops/sync.rs
notoriaga: Oops missed this one. I switched to a log because we mostly seem to be just logging this kind of errors thought the rest of the code. Like in some places we check to see if an object returned from esthri has an etag, and log/skip it if it doesn't. But really there should never be an object without and etag, it's just a quirk of the client we use

crates/esthri/src/ops/sync.rs
silverjam: > Each iteration of the `WalkDir` loop had a potential block (probably by making a call to read_dir). I think you're right in that it's not a big problem because it's all metadata calls not actual io.

My understanding is that it is actual IO, because read_dir would need to make call to the filesystem, which could end up reading from disk (if things aren't already cached somewhere in the kernel) to pull in the directory inode data structure.  The read of the inode is likely small and fast, but it is IO that can potentially block for a long time (if it's a slow drive, or a network drive, etc).

PR #261, opened 2022-02-11
Fix sync up and improve upload tests [DEVINFRA-629]

samvrlewis:
- Upload keys have been randomised to make the tests actually check if
files have been uploaded. Previously these tests were effectively only
testing if the files that had already been uploaded hasn't changed,
which doesn't catch issues with the upload not occurring.

- Fixes the walk dir blocking thread not being spawned as it was being
given an async closure. Also makes directories not get emitted.

### DIFF
--- a/crates/esthri-test/src/lib.rs
+++ b/crates/esthri-test/src/lib.rs
@@ -12,6 +12,7 @@ use tempdir::TempDir;
 
 use esthri::rusoto::*;
 use esthri::*;
+use uuid::Uuid;
 
 pub struct TestGlobal {
     s3client: Option<Arc<S3Client>>,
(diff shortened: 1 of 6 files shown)


crates/esthri-test/src/lib.rs
```diff
@@ -30,6 +31,10 @@ pub fn test_data(name: &str) -> PathBuf {
     test_data_dir().join(name)
 }
 
+pub fn randomised_name(name: &str) -> String {
+    format!("{}-{}", Uuid::new_v4(), name)
```
samvrlewis, line 35: We should probably add a default age out to the esthri-test bucket now, but it'll need to be done a little carefully as some of the download tests rely on files being in the right place, so these shouldn't be deleted. Probably not super high priority as the integration tests don't get run super regularly and don't upload too much, so have created a story to track it: https://swift-nav.atlassian.net/browse/DEVINFRA-630

silverjam (approved):

PR #267, opened 2022-02-22
use system temp dir

notoriaga:
Found a race condition where temp files weren't cleaned up quick enough after a download and some other code that walked the download dir picked up the temp files. This uses the system temp dir instead of the pwd. This also means that should some code panic (or you ctrl-c the bin) the files will still be cleaned up (because the os handles the cleanup now)

### DIFF
--- a/crates/esthri/src/compression.rs
+++ b/crates/esthri/src/compression.rs
@@ -25,17 +25,13 @@ pub const ESTHRI_METADATA_COMPRESS_KEY: &str = "esthri_compress_version";
 
 pub async fn compress_to_tempfile(path: &Path) -> Result<(TempFile, u64)> {
     debug!("compressing: {}", path.display());
-    let dir = match path.parent() {
-        Some(parent) => parent.to_owned(),
-        None => std::env::current_dir()?,
-    };
     let mut src = {
         let f = File::open(path).await?;
         let size = f.metadata().await?.len();
         debug!("old file size: {}", size);
         GzipEncoder::new(BufReader::new(f))
     };
-    let mut dest = TempFile::new(dir, Some(".gz")).await?;
+    let mut dest = TempFile::new(Some(".gz")).await?;
     let new_size = io::copy(&mut src, dest.file_mut()).await?;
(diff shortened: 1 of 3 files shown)


samvrlewis:
Annoyingly `tempfile` has the caveat ["Note: Temporary files cannot be persisted across filesystems."](https://docs.rs/tempfile/latest/tempfile/struct.TempPath.html#method.persist) which is why we chose the `tempfile_in` approach at the time (at the time on CI `/tmp/` was on a different (tmpfs? I think) filesystem so this made the integration tests fail). 

Had a little look and it looks like this is because tempfile uses the `rename` system call https://github.com/Stebalien/tempfile/blob/1a40687e06eb656044e3d2dffa1379f04b3ef3fd/src/file/imp/unix.rs#L131 which doesn't work across filesystems (https://man7.org/linux/man-pages/man2/rename.2.html) - apparently because copying the data is something that should happen in userspace. I wonder if we could/should submit a PR to tempfile to allow persisting via copying/deleting rather than renaming? Otherwise I'm not sure this will be doable as I think it'll stop esthri from being able to download on systems that have `/tmp` tmpfs backed.

notoriaga:
Ahh dang, I knew it couldn't be that easy. I have a fix for this for my use case so it's no big deal, but yeah maybe we can figure something out

notoriaga:
Another way around this might be to make the tmp dir configurable? Maybe something like 

```
ESTHRI_TEMP_DIR=system // what I have here
ESTHRI_TEMP_DIR=dest // what we have now
ESTHRI_TEMP_DIR=/an/actual/path // use that path
``` 

but I also think a contribution to tempfile would be nice

silverjam:
Making things configurable sounds reasonable, I think our ctrl-c handler and panic handler could be improved to clean-up these files too.  I think `rename` is typically used because the operation is atomic at the OS level, other methods are not atomic so you are more likely to have in-between states where things can be half moved.

notoriaga:
closing in favor of https://github.com/swift-nav/esthri/pull/285

PR #270, opened 2022-03-08
Uprev to 8.1.2.

john-michaelburke:


john-michaelburke:
Any chance I could get someone to publish to crates . io as well.

samvrlewis (approved):
I usually do this with `cargo release` directly on master, but I think this should also work as long as you tag the squashed commit after you merge? You should also have permission to push to crates.io, I think, the `swiftnav/devinfra` group is an owner. Let me know if it doesn't work though and I can help.

PR #276, opened 2022-04-04
http server: Uncompress files within archive and use zip compression [DEVINFRA-662] [DEVINFRA-663]

samvrlewis:
Changes the "download archive" functionality of the HTTP server to use
zip compression, rather than tgz, as zip is more widely supported.

Also decompresses files that are compressed within S3 while streaming
out the zip file, so that files are not doubly compressed within
archives.

### DIFF
--- a/crates/esthri-cli/src/http_server.rs
+++ b/crates/esthri-cli/src/http_server.rs
@@ -11,11 +11,11 @@
  */
 
 use std::{
-    borrow::Cow,
     cell::Cell,
     convert::Infallible,
     io::{self, ErrorKind},
     net::SocketAddr,
+    pin::Pin,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc, Mutex,
@@ -23,9 +23,13 @@ use std::{
 };
 
 use anyhow::anyhow;
-use async_compression::tokio::bufread::GzipEncoder;
+
 use async_stream::stream;
-use async_tar::{Builder, Header};
+
+use async_zip::{
+    write::{EntryOptions, ZipFileWriter},
+    Compression,
+};
 use bytes::{Bytes, BytesMut};
 use futures::stream::{Stream, StreamExt, TryStreamExt};
 use hyper::header::CONTENT_ENCODING;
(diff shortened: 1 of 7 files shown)


crates/esthri/src/ops/download.rs
```diff
@@ -67,20 +69,31 @@ where
     download_file(s3, bucket.as_ref(), key.as_ref(), file.as_ref(), true).await
 }
 
-#[logfn(err = "ERROR")]
```
samvrlewis, line 70: I got rid of this because it was causing type errors. I didn't look into it too much but I guess the macro has difficulty parsing the types.

crates/esthri-cli/src/http_server.rs
```diff
@@ -34,12 +38,12 @@ use maud::{html, Markup, DOCTYPE};
 use mime_guess::mime::{APPLICATION_OCTET_STREAM, TEXT_HTML_UTF_8};
 use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
-use sluice::{pipe, pipe::PipeWriter};
```
silverjam, line 37: Can we drop the sluice dep now?

samvrlewis: Yup, have done so now

crates/esthri-cli/tests/integration/http_server.rs
```diff
@@ -1,8 +1,8 @@
+use std::io::Cursor;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::{fs, io};
 
-use flate2::read::GzDecoder;
-use tar::Archive;
```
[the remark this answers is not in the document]
samvrlewis: This one I did already remove - `cargo udeps` claims there's no unused dependencies as well.

crates/esthri/src/ops/download.rs
samvrlewis: I've tricked it into working by having a wrapper function and having the annotation on the wrapper: https://github.com/swift-nav/esthri/pull/276/commits/2247f2763372d571470b9c956d35132cc5d68903 . I don't really understand the original error but this seems to work at least. :man_shrugging:

crates/esthri-cli/tests/integration/http_server.rs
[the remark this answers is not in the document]
silverjam: 🎉

silverjam (approved):

PR #278, opened 2022-04-11
Change default storage class to infrequent access

adrian-kong:
Changes default storage class from STANDARD to STANDARD_IA

DEVINFRA-658

### DIFF
--- a/crates/esthri-cli/src/main.rs
+++ b/crates/esthri-cli/src/main.rs
@@ -128,6 +128,8 @@ enum EsthriCommand {
         key: String,
         /// The path of the local file to read
         file: PathBuf,
+        /// The storage class of file (defaults to STANDARD_IA)
+        storage_class: String
     },
     /// Download an object from S3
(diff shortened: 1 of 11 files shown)


samvrlewis:
It would be nice if this was configurable, so users could choose the storage class they want to upload with. Some consumers might have S3 access patterns where standard or another storage class might be more appropriate than IA.

What do you think about doing something like:

```
enum S3StorageClass {
    Standard,
    Standard_IA,
    // ..other classes..
}
```

And then to avoid breaking the API we could get the existing upload functions to upload with a default class but add new upload methods like:

```
pub async fn upload_with_storage_class<T>(
    s3: &T,
    bucket: impl AsRef<str>,
    key: impl AsRef<str>,
    file: impl AsRef<Path>,
    storage_class: S3StorageClass
)
```

Probably should allow it to be configurable from the CLI too, but that could probably be a separate task if this task is getting too big.

PR #284, opened 2022-04-21
Added CLI for storage class

adrian-kong:
Added --storage ... args for put command, defaults to STANDARD_IA
Logging info outputs storage_class as well

### DIFF
--- a/crates/esthri-cli/src/main.rs
+++ b/crates/esthri-cli/src/main.rs
@@ -128,7 +128,7 @@ enum EsthriCommand {
         #[structopt(long)]
         key: String,
         /// The storage class of the object (example: RRS)
-        #[structopt(long = "storage", default_value = "STANDARDIA")]
+        #[structopt(long = "storage", default_value = "S3StorageClass::Standard")]
         storage_class: String,
         /// The path of the local file to read
         file: PathBuf,
@@ -387,8 +387,8 @@ async fn dispatch_esthri_cli(cmd: EsthriCommand, s3: &S3Client) -> Result<()> {
             ref storage_class,
             compress,
         } => {
-            let storage_class = S3StorageClass::from_str(storage_class.as_ref())
-                .unwrap_or(S3StorageClass::StandardIA);
+            let storage_class =
+                From::from_str(storage_class.as_ref()).unwrap_or(S3StorageClass::StandardIA);
 
             if compress {
                 esthri::upload_compressed_with_storage_class(s3, bucket, key, file, storage_class)
--- a/crates/esthri-cli/src/main.rs
+++ b/crates/esthri-cli/src/main.rs
@@ -13,10 +13,8 @@
 
 mod http_server;
 
-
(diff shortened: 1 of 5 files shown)


crates/esthri-cli/src/main.rs
```diff
@@ -126,6 +127,9 @@ enum EsthriCommand {
         /// The key name of the object (example: a/key/name.bin)
         #[structopt(long)]
         key: String,
+        /// The storage class of the object (example: RRS)
+        #[structopt(long = "storage", default_value = "STANDARDIA")]
+        storage_class: String,
```
notoriaga, line 132: `clap` automatically supports types that implement "FromStr", so instead of a `String` here and parsing it later, you can do

```rust
#[structopt(long = "storage", default_value = "S3StorageClass::StandardIA")]
storage_class: S3StorageClass,
```

silverjam:
> Added --storage ... args for put command, defaults to STANDARD_IA Logging info outputs storage_class as well

Default should remain `STANDARD` but we'll need an easy way to control what the default is (via an environment VAR) so that we can tweak that in SITL.

crates/esthri-cli/src/main.rs
```diff
@@ -13,8 +13,10 @@
 
 mod http_server;
 
+
 use std::env;
 use std::ffi::OsStr;
+
```
notoriaga, line 19: nit: whitespace

```suggestion
use std::env;
use std::ffi::OsStr;
```

Generally my preference for organizing deps is 

```
<std>
<blank>
<third party>
<blank>
<local>
```

with no new lines between deps in the same category

notoriaga: (this is not a hard and fast rule at all and is unenforced by rust-format unless you use a nightly compiler in which case you can tell it to do this)

crates/esthri/src/rusoto.rs
```diff
@@ -143,10 +159,11 @@ impl FromStr for S3StorageClass {
             "STANDARD_IA" => Ok(S3StorageClass::StandardIA),
             "INTELLIGENT_TIERING" => Ok(S3StorageClass::IntelligentTiering),
             "ONEZONE_IA" => Ok(S3StorageClass::OneZoneIA),
-            "GLACIAL_INSTANT_RETRIEVAL" => Ok(S3StorageClass::GlacialInstantRetrieval),
-            "GLACIAL_FLEXIBLE_RETRIEVAL" => Ok(S3StorageClass::GlacialFlexibleRetrieval),
-            "GLACIAL_DEEP_ARCHIVE" => Ok(S3StorageClass::GlacialDeepArchive),
+            "GLACIER_INSTANT_RETRIEVAL" => Ok(S3StorageClass::GlacialInstantRetrieval),
```
notoriaga, line 162: are these renames from `GLACIAL` -> `GLACIER` intended?

notoriaga (approved):

PR #285, opened 2022-04-22
Added temp file new with directory using config

adrian-kong:
Temp file directory using config 

need to fix up the tests but would this be basically the idea

### DIFF
--- a/crates/esthri/src/compression.rs
+++ b/crates/esthri/src/compression.rs
@@ -19,19 +19,24 @@ use tokio::{
     io::{self, BufReader},
 };
 
-use crate::{tempfile::TempFile, Result};
+use crate::{tempfile::TempFile, Config, Result};
 
 pub const ESTHRI_METADATA_COMPRESS_KEY: &str = "esthri_compress_version";
 
 pub async fn compress_to_tempfile(path: &Path) -> Result<(TempFile, u64)> {
     debug!("compressing: {}", path.display());
+    let dir = match (Config::global().temp_dir_path(), path.parent()) {
+        (Some(dir), _) => dir,
+        (None, Some(parent)) => parent.to_owned(),
+        _ => std::env::current_dir()?,
+    };
     let mut src = {
         let f = File::open(path).await?;
         let size = f.metadata().await?.len();
         debug!("old file size: {}", size);
         GzipEncoder::new(BufReader::new(f))
     };
-    let mut dest = TempFile::new(Some(".gz")).await?;
+    let mut dest = TempFile::new(dir, Some(".gz")).await?;
     let new_size = io::copy(&mut src, dest.file_mut()).await?;
     dest.rewind().await?;
     debug!("new file size: {}", new_size);
--- a/crates/esthri/src/compression.rs
+++ b/crates/esthri/src/compression.rs
@@ -25,17 +25,13 @@ pub const ESTHRI_METADATA_COMPRESS_KEY: &str = "esthri_compress_version";
 
 pub async fn compress_to_tempfile(path: &Path) -> Result<(TempFile, u64)> {
     debug!("compressing: {}", path.display());
-    let dir = match path.parent() {
-        Some(parent) => parent.to_owned(),
-        None => std::env::current_dir()?,
-    };
     let mut src = {
         let f = File::open(path).await?;
         let size = f.metadata().await?.len();
         debug!("old file size: {}", size);
         GzipEncoder::new(BufReader::new(f))
     };
-    let mut dest = TempFile::new(dir, Some(".gz")).await?;
+    let mut dest = TempFile::new(Some(".gz")).await?;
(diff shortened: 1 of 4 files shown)


crates/esthri/src/ops/download.rs
```diff
@@ -132,7 +132,7 @@ where
         obj_info.parts * obj_info.size as u64
     );
     let dir = init_download_dir(download_path).await?;
-    let mut dest = TempFile::new(dir, None).await?;
+    let mut dest = TempFile::new_with_dir(dir, None).await?;
```
notoriaga, line 135: so I think the behavior we want is 

- if ESTHRI_TEMP_DIR is unset, we create temp file in the same directory where we are downloading to.
- if ESTHRI_TEMP_DIR is set, write temp files there.

so you might wanna make the field in the config struct optional

```rust
struct Config {
    temp_dir_path: Option<TempDirPath>,
}
```

and then do

```rust
    let dir = init_download_dir(download_path).await?;
    let mut dest = if let Some(dir) = Config::global().temp_dir_path() {
        TempFile::new(dir, None).await?
    } else {
        TempFile::new(dir, None).await?
    };
```

crates/esthri/src/compression.rs
```diff
@@ -25,17 +25,13 @@ pub const ESTHRI_METADATA_COMPRESS_KEY: &str = "esthri_compress_version";
 
 pub async fn compress_to_tempfile(path: &Path) -> Result<(TempFile, u64)> {
     debug!("compressing: {}", path.display());
-    let dir = match path.parent() {
```
notoriaga, line 28: similarly, this should probably be 

```rust
    let dir = match (Config::global().temp_dir_path(), path.parent()) {
        (Some(dir), _) => dir,
        (None, Some(parent)) => parent.to_owned(),
        _ => std::env::current_dir()?,
    };
```

crates/esthri/src/tempfile.rs
```diff
@@ -19,7 +19,7 @@ pub struct TempFile {
 pub const TEMP_FILE_PREFIX: &str = ".esthri_temp";
 
 impl TempFile {
-    pub async fn new(dir: PathBuf, suffix: Option<&str>) -> Result<Self> {
+    pub async fn new_with_dir(dir: PathBuf, suffix: Option<&str>) -> Result<Self> {
```
notoriaga, line 22: with those changes I mentioned I think you should be able to get away without making any changes to this file

crates/esthri/src/compression.rs
notoriaga: i.e. prefer writing to ESTHRI_TEMP_DIR if its set, then prefer creating the temp file next to the original, finally default to current_dir (which realistically shouldn't happen because the parent directory should always be set)

crates/esthri/src/config.rs
```diff
@@ -174,13 +183,20 @@ impl Config {
...
-    /// from awscli, changing this size will affect the calculation of ETags.  Defaults to
-    /// [STORAGE_CLASS].
+    /// The default storage class for put requests to S3. STANDARD is the default storage class
+    /// unless specified. View AWS documentation for more specifications on other storage classes.
+    /// Defaults to [STORAGE_CLASS].
     pub fn storage_class(&self) -> S3StorageClass {
         self.storage_class.0
     }
 
+    /// The default size of parts in a multipart upload to S3.  8 MiB is the default chunk size
```
notoriaga, line 193: looks great! I think once you update this comment we should be good to go

notoriaga:
so I tried this and it seems to work - one thing I noticed is that if the path doesn't exist you get an error. I think that's fine but we could also create the directory if it doesn't exist

notoriaga (approved):

crates/esthri/src/tempfile.rs
```diff
@@ -21,6 +22,9 @@ pub const TEMP_FILE_PREFIX: &str = ".esthri_temp";
 impl TempFile {
     pub async fn new(dir: PathBuf, suffix: Option<&str>) -> Result<Self> {
         let suffix = suffix.unwrap_or_default().to_owned();
+        if !&dir.exists() {
+            fs::create_dir_all(&dir)?;
+        }
```
notoriaga, line 27: should be 

```rust
tokio::fs::create_dir_all(&dir).await?;
```

in an async context you want to stick to async versions of functions if they exist. a blocking call inside an async runtime can gum up the executor which can manifest as a performance hit or deadlock. for something like this it's not a huge deal because I doubt that function blocks for long but its an easy fix.

and I think you should be able to drop the `&` in `!&dir.exists()` - it should just return a bool

notoriaga: the reason for this is that `async` in rust implements [cooperative multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking) - so it's up to author to make sure nothing takes too much cpu time before yielding to other tasks

notoriaga (approved):

PR #302, opened 2022-05-28
add get_bucket_location

notoriaga:


crates/esthri/src/rusoto.rs
```diff
@@ -284,3 +285,21 @@ where
     .await
     .map_err(Error::CreateMultipartUploadFailed)
 }
+
+pub async fn get_bucket_location<T>(s3: &T, bucket: &str) -> Result<String>
+where
+    T: S3,
+{
+    let region = handle_dispatch_error(|| {
```
john-michaelburke, line 293: We use rusoto core Region object to attempt to grab from a series of default locations. Not sure if that would be of any use.

https://docs.rs/rusoto_core/latest/rusoto_core/enum.Region.html

john-michaelburke (approved):

crates/esthri/src/rusoto.rs
notoriaga: We use str as the region in the rest of the api, so I feel like that's something we should change all at once if we want

john-michaelburke (approved):

john-michaelburke:
I guess we could use any region to get the client, discern the bucket's, then grab the correct client 🤷

PR #303, opened 2022-05-30
Build for mac and winodws [DEVINFRA-768]

adrian-kong:
Builds other releases


.github/workflows/main.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Build
+        shell: bash
+        run: |
+          cargo make --profile ${{ matrix.profile }} build
+
+#      - name: Package Release
+#        shell: bash
+#        run: |
+#          touch LICENSE.txt
+#          cargo make zip-release-binary-for-target \
```
adrian-kong: using https://github.com/sagiegurari/cargo-make/blob/master/docs/cargo_make_task_list.md
but it gets skipped locally - could just upload the esthri file?

Makefile.toml
```diff
@@ -14,6 +14,14 @@ BUILD_FLAVOR = "--release"
 BUILD_FLAVOR = "--release"
 TARGET = "--target=x86_64-unknown-linux-musl"
 
+[env.'release+mac']
+BUILD_FLAVOR = "--release"
+TARGET = "--target=x86_64-apple-darwin"
```
adrian-kong, line 19: should use `x86_64` instead of `aarch64`?

samvrlewis: Both, if possible, so we can support both x86 and M1 macs

.github/workflows/main.yml
samvrlewis: You could try to figure out why its getting skipped (suspect its something to do with the multi-crate setup) with `cargo make --verbose zip-release-binary-for-target` or maybe, yeah, we could just add our own commands to build + zip

adrian-kong:
reopened as #305 because of some error i had `dst refspec refs matches more than one`

PR #305, opened 2022-05-31
Build for mac and winodws [DEVINFRA-768]

adrian-kong:
Builds mac aarch64, x86_64 and windows releases

PR #307, opened 2022-06-01
Removing CommandExt

adrian-kong:
CommandExt only supports unix:
https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html

Building on windows fails in #305, needs to use windows...
https://doc.rust-lang.org/std/os/windows/process/trait.CommandExt.html

so maybe i took out the commandext overall but the output does not return success code of msg so might have to remove `.success()` in all the cli tests?

### DIFF
--- a/crates/esthri-cli/src/main.rs
+++ b/crates/esthri-cli/src/main.rs
@@ -227,10 +227,10 @@ fn call_real_aws() {
     let aws_tool_path = env::var(REAL_AWS_EXECUTABLE_ENV_NAME)
         .unwrap_or_else(|_| REAL_AWS_EXECUTABLE_DEFAULT.to_string());
 
-    Command::new(&aws_tool_path)
+    let status = Command::new(&aws_tool_path)
         .args(args)
         .stdout(Stdio::inherit())
-        .output()
+        .status()
         .expect(
             format!(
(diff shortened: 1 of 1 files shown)


crates/esthri-cli/src/main.rs
```diff
@@ -228,12 +227,17 @@ fn call_real_aws() {
...
+        .args(args)
+        .stdout(Stdio::inherit())
+        .output()
+        .expect(
+            format!(
+                "Executing aws didn't work. Is it installed and available as {:?} ",
+                aws_tool_path
+            )
+            .as_str(),
+        );
```
notoriaga, line 240: I think you can do something like 

```suggestion
    let status = Command::new(&aws_tool_path)
        .args(args)
        .stdout(Stdio::inherit())
        .status()
        .expect(
            format!(
                "Executing aws didn't work. Is it installed and available as {:?} ",
                aws_tool_path
            )
            .as_str(),
        );
    std::process::exit(status.code().unwrap_or(-1));
```

to propagate the status code

notoriaga: the struct returned by `.output()` also has a `.status()` method which would get you the same thing, but I think status is better here because we are not interested in the output (as in stdout/err)

Issue #277, opened 2022-04-11
esthri-cli isnt published on crates.io


jayvdb:

jeffbski-sketch:
Makes it much easier to install without downloading the git repo

jayvdb:
Yes, it will mean it can be installed using https://crates.io/crates/cargo-quickinstall or https://crates.io/crates/cargo-binstall for very quick prebuilt binary installs, perfect for CI.

john-michaelburke:
https://crates.io/crates/esthri-cli

PR #335, opened 2022-08-29
Allow esthri-cli crate to be published.

john-michaelburke:
Not entirely sure if the release.toml in the root will handle both crates or if we need to provide a separate release.toml for both? Seems like everything is abstracted though so maybe this is fine.

fixes https://github.com/swift-nav/esthri/issues/277

### DIFF
--- a/crates/esthri-cli/release.toml
+++ b/crates/esthri-cli/release.toml
@@ -1,1 +1,0 @@
-publish = false


notoriaga:
I think a `cargo release --workspace --exclude esthri-test` from the root of the project should work (should be equivalent to running it on each package).

john-michaelburke:
> I think a `cargo release --workspace --exclude esthri-test` from the root of the project should work (should be equivalent to running it on each package).

Well there is also a release.toml file in the esthri-test crate would that sufficient or would --workspace ignore it?

notoriaga:
> > I think a `cargo release --workspace --exclude esthri-test` from the root of the project should work (should be equivalent to running it on each package).
> 
> Well there is also a release.toml file with `publish=false` in the esthri-test crate would that be sufficient or would --workspace ignore it?

Ah nice yeah should be fine without the flag then

silverjam (approved):

PR #347, opened 2022-09-06
Only release on tagged branch

adrian-kong:
I believe this causes fake releases as per https://github.com/swift-nav/package-registry/pull/63, 
Fixed in https://github.com/swift-nav/ICBINS/commit/8f2ff3bfc214e90be470a282c1176ed1464a227f for ICBINS

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,9 +4,8 @@ name: Build macOS and Windows
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
+  # Triggers the workflow on release tags
   push:
-    branches: [ main ]
     tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -4,9 +4,8 @@ name: Build macOS and Windows
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
+  # Triggers the workflow on release tags
```
adrian-kong, line 7: It doesn't run this workflow in this repo for some reason... ICBINS has issues with this line.

All the other repos release without branch specifier
```
name: Publish
on:
  push:
    tags:
      - "*"
``` 

Maybe something to do with the Jenkins

silverjam (approved):

PR #349, opened 2022-09-08
esthri-cli: serve: add support for configuring a list of allowed prefixes

silverjam:
Adds a list "directory" prefixes to allow, for example:

```
cargo run -- serve --bucket esthri-test --allowed-prefix test_upload/ --allowed-prefix test_list_directory/
```

### DIFF
--- a/crates/esthri-cli/src/http_server.rs
+++ b/crates/esthri-cli/src/http_server.rs
@@ -215,6 +215,13 @@ fn with_bucket(bucket: String) -> impl Filter<Extract = (String,), Error = Infal
     warp::any().map(move || bucket.clone())
 }
 
+fn with_allowed_prefixes(
+    allowed_prefixes: &[String],
+) -> impl Filter<Extract = (Vec<String>,), Error = Infallible> + Clone {
+    let allowed_prefixes = allowed_prefixes.to_vec();
+    warp::any().map(move || allowed_prefixes.clone())
+}
+
 fn with_s3_client(
     s3_client: S3Client,
 ) -> impl Filter<Extract = (S3Client,), Error = Infallible> + Clone {
@@ -224,10 +231,12 @@ fn with_s3_client(
 pub fn esthri_filter(
     s3_client: S3Client,
     bucket: &str,
+    allowed_prefixes: &[String],
 ) -> impl Filter<Extract = (http::Response<Body>,), Error = warp::Rejection> + Clone {
     warp::path::full()
         .and(with_s3_client(s3_client))
         .and(with_bucket(bucket.to_owned()))
+        .and(with_allowed_prefixes(allowed_prefixes))
         .and(warp::query::<DownloadParams>())
         .and(warp::header::optional::<String>("if-none-match"))
         .and_then(download)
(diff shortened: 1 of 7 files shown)


crates/esthri-cli/src/main.rs
```diff
@@ -231,13 +233,12 @@ fn call_real_aws() {
         .args(args)
         .stdout(Stdio::inherit())
         .status()
-        .expect(
-            format!(
+        .unwrap_or_else(|_| {
```
silverjam, line 236: clippy fix-up

john-michaelburke:
This is great. I think a useful addition, maybe in another PR, to this would be a tailored landing page if a user reaches a non-whitelisted path. Not sure if that would be something best handled here or by the esthri consumer.

crates/esthri-cli/src/http_server.rs
```diff
@@ -224,10 +231,12 @@ fn with_s3_client(
...
     bucket: &str,
+    allowed_prefixes: &[String],
 ) -> impl Filter<Extract = (http::Response<Body>,), Error = warp::Rejection> + Clone {
     warp::path::full()
         .and(with_s3_client(s3_client))
         .and(with_bucket(bucket.to_owned()))
+        .and(with_allowed_prefixes(allowed_prefixes))
         .and(warp::query::<DownloadParams>())
         .and(warp::header::optional::<String>("if-none-match"))
         .and_then(download)
```
notoriaga, line 242: Is it possible to do something like 

```rust
let path = warp::path::full()
    .and(with_s3_client(s3_client))        
    .and(with_bucket(bucket.to_owned()))
    .and(with_allowed_prefixes(allowed_prefixes))
    .and(warp::query::<DownloadParams>())
    .and(warp::query::<DownloadParams>())
    .and(warp::header::optional::<String>("if-none-match"))
    .and(warp::header::optional::<String>("if-none-match"))
    .and_then(download)
    .and_then(download)
if !allowed_prefixes.is_empty() {
    path.and(with_allowed_prefixes(allowed_prefixes));
} else {
    path
}
```
then the filter doesn't have to worry about an empty prefix list

silverjam: First thought it probably not?  This creates a function signature, so if we conditionally add it, then it would produce two diferent function signatures.

silverjam: I was able to do some preprocessing in the filter setup, which makes the code a little clearer

crates/esthri-cli/tests/integration/http_server.rs
```diff
@@ -40,6 +41,88 @@ async fn test_fetch_object() {
     assert_eq!(body, "this file has contents\n");
 }
 
+#[tokio::test]
```
achubbic-snav, line 44: Please add comment describing what the test aims to validate, including whether this addresses nominal behavior or edge cases if applicable.

achubbic-snav (approved):
Marking as approved and trusting @notoriaga 's vigilant eye- current knowledge of filters insufficient.

crates/esthri-cli/tests/integration/http_server.rs
silverjam: done

PR #351, opened 2022-09-10
add sync delete for local to remote [DEVINFRA-900] [DEVINFRA-13]

silverjam:
+ Added `delete` and `delete_streaming` API calls to support the new `sync --delete` feature

+ Add `sync --delete` for local to remote sync direction

+ Splits out bits from `esthri` lib that `esthri-test` needs so that there is no longer a recursive dependency between the libraries, this was causing `rust-analyzer` to fail on VSCode

+ Don't build `release` in CI it should only be necessary for publishing the `esthri` CLI tool and it'll greatly speed up CI builds

+ Enable more optimizations for release builds since they're now only done for publish


crates/esthri/src/ops/sync.rs
```diff
@@ -445,10 +449,51 @@ where
...
+                    let obj_path = String::from(key) + &s3_metadata.s3_suffix;
+                    Ok(Some(obj_path))
+                } else {
+                    Err(err.into())
+                }
+            } else {
+                Ok(None)
+            }
+        } else {
+            Err(Error::MetadataNone)
```
silverjam, line 489: Not sure when this would happen

crates/esthri/src/ops/sync.rs
```diff
@@ -445,10 +449,51 @@ where
 
     sync_tasks
         .buffer_unordered(task_count)
-        .try_collect()
+        .try_for_each_concurrent(task_count, |_| future::ready(Ok(())))
```
silverjam, line 452: Thinking that this will improve memory usage of the sync process since it does not collection result into a vector (which is immediately thrown away anyway).

crates/esthri/src/ops/sync.rs
```diff
@@ -710,7 +755,7 @@ where
...
```
silverjam, line 758: Again, hoping this will improve memory usage

release.toml
```diff
@@ -1,7 +1,7 @@
-pre-release-commit-message = "Release: {{version}}"
 post-release-commit-message = "Start next development iteration: {{next_version}}"
-tag-message = "{{tag_name}}"
+pre-release-commit-message = "release: {{crate_name}}: version {{version}}"
```
silverjam, line 2: This is mostly stealing the config from Swift CLI which has been working well, this config has had a few issues in the past

silverjam (commented):
Note this is just the local to remote direction-- remote to local will be implemented in another pull request.

cc @achubbic-snav

crates/esthri/src/ops/sync.rs
```diff
@@ -445,10 +449,51 @@ where
...
+    bucket: &str,
+    key: &str,
+    directory: &Path,
+    filters: &[GlobFilter],
+    task_count: usize,
+) -> Result<()>
+where
+    T: S3 + Send + Sync + Clone,
+{
+    let stream = flattened_object_listing(s3, bucket, key, directory, filters, false);
```
john-michaelburke, line 473: This touches on one of the questions that came up in the PP session we had with @achubbic-snav which was how the glob filters should be handled with respect to the "delete" feature of sync. I believe here it is filtering what will be synced up ( include/exclude ) but also filtering what would be considered for deletion in the destination. Maybe I'm overthinking this.

```
include: *.txt
exclude: *
```

src/
  - asdf.txt
  - asdf2.txt
  - asdf3.mpg

dest/
  - asdf.txt
  - asdf2.txt
  - asdf3.mpg
  - asdf4.mpg

In this example due to the filters applied on both src and dest, I believe the `asdf4.mpg` would not be considered for deletion.

john-michaelburke (approved):

release.toml
silverjam: It will create a commit for each crate and then publish them in the correct order according to who's dependent on whom.

crates/esthri/src/ops/sync.rs
silverjam: We tested this with the aws cli and the filters applied to both upload/download and what was considered for deletion-- which coincidentally was the easiest thing to implement too.

john-michaelburke: Right on. Interesting chicken and the egg problem.

achubbic-snav (approved):

crates/esthri/src/ops/delete.rs
```diff
@@ -0,0 +1,35 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+use log_derive::logfn;
+
+use crate::errors::Result;
+use crate::rusoto::{DeleteObjectRequest, S3};
+
+#[logfn(err = "ERROR")]
+pub async fn delete<T>(s3: &T, bucket: impl AsRef<str>, key: impl AsRef<str>) -> Result<()>
```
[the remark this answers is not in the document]
silverjam: This is already pretty effectively handled by the stdlib, so I don't think we need a wrapper unless there's some part of our abstractions that would benefit from having a local wrapper

Makefile.toml
```diff
@@ -3,28 +3,41 @@ default_to_workspace = false
...
 [env]
 BUILD_FLAVOR = ""
+VERBOSE = ""
 TLS_PROVIDER = "rustls"
 TARGET = ""
 WORKSPACE = "${CARGO_MAKE_WORKING_DIRECTORY}"
 
 [env.release]
 BUILD_FLAVOR = "--release"
+RUSTFLAGS="-C link-arg=-fuse-ld=lld -C codegen-units=1 -C panic=abort -C lto=fat -C opt-level=3"
```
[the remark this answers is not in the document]
silverjam: These should probably be moved to a Cargo.toml profile at some point, but right now it's just used to build the release artifacts for the `esthri` CLI that we publish.  The things I didn't enable on Windows are things I didn't think would work there, Windows support for Esthri is still pretty iffy, so could definitely use some more love later.

crates/esthri/tests/integration/sync_test.rs
```diff
@@ -96,6 +99,7 @@ fn test_sync_up_without_slash() {
         destination,
         filters.as_deref(),
         false,
+        false,
```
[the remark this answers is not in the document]
silverjam: Before we publish the next release I think we should go through and wrap up some of these options in to struct, it'll improve out API stability-- right now we'll need another major version bump to add these bools to this API so if we're going to another major version bump we should instead add some facilities that'll hopefully avoid another one in the future. cc @achubbic-snav

crates/esthri/src/ops/delete.rs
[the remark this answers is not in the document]
john-michaelburke: I think I had mentioned this before reading this was only implementing delete for local to remote. What I was proposing is that we abstract delete functionality regardless if it was deleting local to remote or remote to local. I assume that would be handled in another PR abstracted somewhere close to the top level sync function.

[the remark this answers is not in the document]
silverjam: 👍 -- yes, indeed it is

PR #363, opened 2022-09-27
Adding panic test for unset environment variable

gborough:
Purpose:
Part of onboarding exercise: https://swift-nav.atlassian.net/browse/DOL-160, please feel free to reject this PR

Technical Detail:
By setting and unsetting environment variable for the "WORKSPACE" key, it's able to test whether the unwrapping of test_data_dir() would panic when unit testing

### DIFF
--- a/crates/esthri-test/src/lib.rs
+++ b/crates/esthri-test/src/lib.rs
@@ -133,3 +133,22 @@ fn init_s3client() {
         Some(_) => {}
     }
 }
+
+#[cfg(test)]
+mod panic_test {
+    use super::*;
+
+    #[test]
+    #[should_panic]
+    fn test_dir_unset() {
+        let key = "WORKSPACE";
+        std::env::set_var(key, "somedir");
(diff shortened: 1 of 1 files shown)


adrian-kong (commented):
Don't think it's necessary to test a test function since `WORKSPACE` is set by default in `Makefile.toml`
https://github.com/swift-nav/esthri/blob/9333d996d855f7baf5abc466db2a613a60e70407/Makefile.toml#L9
So it shouldn't ever be unset I think

adrian-kong:
Could add some test cases for missing S3StorageClass uploads (right now only tests Standard and RRS, could generalize into a S3StorageClass test or something) if needing some ideas...
https://github.com/swift-nav/esthri/blob/c0267742665e6ea890d8e96af3e14837b29a30f5/crates/esthri/tests/integration/upload_test.rs#L119-L163

gborough:
Thanks for the idea, I will close this PR and work on your suggestions

PR #375, opened 2022-10-10
server shutdown fix

notoriaga:
Having some trouble shutting down the server when running in k8s. This prevents two ctrl-c handlers from being set up while in server mode (not sure this is the issue) and cleans up the termination code a bit (definitely not the issue)

### DIFF
--- a/crates/esthri-cli/src/http_server.rs
+++ b/crates/esthri-cli/src/http_server.rs
@@ -11,7 +11,6 @@
  */
 
 use std::{
-    cell::Cell,
     convert::Infallible,
     io::{self, ErrorKind},
     net::SocketAddr,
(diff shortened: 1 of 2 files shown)


silverjam (approved):

achubbic-snav (approved):

silverjam:
Why do we need to bump to 1.64?

notoriaga:
> Why do we need to bump to 1.64?

I bumped for const mutex and just picked the latest because I didn't know offhand when it was stabilized. For some reason I thought we had a "server" feature or something so that it wouldn't require a MSRV bump for library users of this crate but I guess not

silverjam:
superseded by #377

PR #377, opened 2022-10-11
http server shutdown fix-ups [DEVINFRA-959]

silverjam:
from #375 
> Having some trouble shutting down the server when running in k8s. This prevents two ctrl-c handlers from being set up while in server mode (not sure this is the issue) and cleans up the termination code a bit (definitely not the issue)

implemented graceful shutdown SIGTERM handler shown here: https://stackoverflow.com/a/71279547/749342

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -7,6 +7,7 @@ VERBOSE = ""
 TLS_PROVIDER = "rustls"
 TARGET = ""
 WORKSPACE = "${CARGO_MAKE_WORKING_DIRECTORY}"
+MSRV = ""
 
 [env.release]
 BUILD_FLAVOR = "--release"
@@ -39,6 +40,15 @@ VERBOSE = "-vv"
 TLS_PROVIDER = "nativetls"
 VERBOSE = "-vv"
 
+[env.'dev+msrv']
+MSRV = "+1.56.1"
+VERBOSE = "-vv"
+
+[env.'dev+msrv+nativetls']
+TLS_PROVIDER = "nativetls"
+MSRV = "+1.56.1"
+VERBOSE = "-vv"
+
 [tasks.check]
 command = "cargo"
 args = [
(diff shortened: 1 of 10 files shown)


crates/esthri/src/ops/upload.rs
```diff
@@ -529,7 +529,11 @@ fn upload_part_requests<'a>(
 ) -> impl Stream<Item = UploadPartRequest> + 'a {
     let last_part = {
         let remaining = file_size % part_size;
-        (remaining > 0).then(|| (file_size / part_size + 1, remaining))
+        if remaining > 0 {
```
silverjam, line 532: Required for MSRV of Rust 1.56.1

crates/esthri-internals/Cargo.toml
```diff
@@ -9,11 +9,13 @@ repository = "https://github.com/swift-nav/esthri/"
 homepage = "https://github.com/swift-nav/esthri/"
 readme = "../../README.md"
 keywords = ["aws", "s3"]
+rust-version = "1.56.1"
```
silverjam, line 12: If we want to support earlier rust versions, we'd need to remove this key

silverjam (commented):
I used the tool `cargo-msrv` to validate that 1.56.1 is in fact the lowest version we support: https://crates.io/crates/cargo-msrv

Makefile.toml
```diff
@@ -7,22 +7,23 @@ VERBOSE = ""
 TLS_PROVIDER = "rustls"
 TARGET = ""
 WORKSPACE = "${CARGO_MAKE_WORKING_DIRECTORY}"
+MSRV = ""
 
 [env.release]
 BUILD_FLAVOR = "--release"
-RUSTFLAGS="-C link-arg=-fuse-ld=lld -C codegen-units=1 -C panic=abort -C lto=fat -C opt-level=3"
```
gborough, line 13: Maybe in the future we can experiment with using the Mold linker if we really care about linking performance

notoriaga:
> I used the tool `cargo-msrv` to validate that 1.56.1 is in fact the lowest version we support: https://crates.io/crates/cargo-msrv

very cool!

Makefile.toml
notoriaga: I've been using it for a while, makes a big difference

notoriaga (approved):

crates/esthri/Cargo.toml
```diff
@@ -9,6 +9,7 @@ repository = "https://github.com/swift-nav/esthri/"
 homepage = "https://github.com/swift-nav/esthri/"
 readme = "../../README.md"
 keywords = ["aws", "s3"]
+rust-version = "1.56.1"
```
gborough, line 12: we can also create a .clippy.toml file at the project root dir that specifies msrv = "1.56.1"

gborough (approved):

crates/esthri/Cargo.toml
silverjam: Cool, feel free to submit a PR for this :-)

PR #379, opened 2022-10-12
Publish arm64 binaries

joelynch:


silverjam:
@joelynch not sure why this is marked failure, it looks like it's ready

joelynch:
Thanks for working through all those cc & linking issues @silverjam! I was going to try using [cross](https://github.com/cross-rs/cross) but this looks great.

silverjam (approved):

Makefile.toml
```diff
@@ -20,12 +20,28 @@ TARGET = "--target=x86_64-unknown-linux-musl"
 RUSTFLAGS="-C codegen-units=1 -C panic=abort -C lto=fat -C opt-level=3"
 VERBOSE = "-vv"
 
+[env.'release+static+aarch64']
+CC_aarch64_unknown_linux_musl="clang"
+AR_aarch64_unknown_linux_musl="llvm-ar"
+CFLAGS_aarch64_unknown_linux_musl="-I/usr/aarch64-linux-gnu/include"
```
isaactorz, line 26: Is this ok?

I'd think for a musl build we'd want to point to something like `/usr/aarch64-linux-musl/include`.

joelynch: I tested this one out locally and it does work in a linux docker container which might mean that this flag is being ignored.

isaactorz: If you run file or ldd commands against the binary that should tell you if it was fully statically linked (uses musl libc)

silverjam: I think these are libc independent headers?  It wasn't needed for me on my Linux machine, but didn't work in CI without it.  This should be part of the "sysroot" which should be discovered automatically, but isn't for some reason on GitHub actions.

silverjam: Does a musl variant of that path even exist?

silverjam: > I tested this one out locally and it does work in a linux docker container which might mean that this flag is being ignored.

It's definitely not ignored, with verbose output you can see the flag being included in compile commands.

silverjam: > I think these are libc independent headers?  It wasn't needed for me on my Linux machine, but didn't work in CI without it.  This should be part of the "sysroot" which should be discovered automatically, but isn't for some reason on GitHub actions.

Maybe not libc independent, but there's no equivalent package for musl cross compilation: https://packages.ubuntu.com/bionic/libc6-dev-arm64-cross so I don't know why it works on my machine.

silverjam: Most of this is from https://github.com/briansmith/ring/issues/1414#issuecomment-1055177218 but looks like ring itself doesn't have issues with the sysroot, so not sure why we do

isaactorz: 🤔 . I do have a musl variant on my path at `/usr/include/x86_64-linux-musl/`. (not cross compiling obv).

I guess I was just concerned that if those flags were necessary you might've not actually been linking musl libc, but joe tested the binary and it's fully statically linked.

silverjam: > 🤔 . I do have a musl variant on my path at `/usr/include/x86_64-linux-musl/`. (not cross compiling obv).
> 

Right, but there doesn't appear to be a aarch64 variant of this path

> 
> 
> I guess I was just concerned that if those flags were necessary you might've not actually been linking musl libc, but joe tested the binary and it's fully statically linked.

Could strace the whole compile process locally to figure out where the "bits" files are coming from

isaactorz (approved):

silverjam:
To feed a fed 🐴 ...

On my system strace reveals that the following include directories are discovered/used with the CFLAG var set:

```
❯ grep 'open.*usr.*aarch64.*include.*' strace.txt | cut -d' ' -f3- | sed 's@.*"\(.*\)".*@\1@' | sort | uniq

/usr/aarch64-linux-gnu/include/bits/libc-header-start.h
/usr/aarch64-linux-gnu/include/bits/long-double.h
...
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/alloca.h
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/arm_neon.h
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/asm-generic/errno-base.h
...
```

Without the CFLAG var set, it only uses the `.../gcc-cross/...` style paths:

```
❯ grep 'open.*usr.*aarch64.*include.*' strace-no-cflags.txt | cut -d' ' -f3- | sed 's@.*"\(.*\)".*@\1@' | sort | uniq

/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/alloca.h
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/arm_neon.h
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/asm-generic/errno-base.h
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/asm-generic/errno.h
/usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/asm/errno.h
...
```

Interestingly, if you inspect one of these files, they resolve to the same manually specified path in the CFLAGS var:

```
❯ realpath /usr/bin/../lib/gcc-cross/aarch64-linux-gnu/9/../../../../aarch64-linux-gnu/include/alloca.h

/usr/aarch64-linux-gnu/include/alloca.h
```

This comes from the libc6.*cross package:

```
❯ dpkg -S /usr/aarch64-linux-gnu/include/alloca.h
libc6-dev-arm64-cross: /usr/aarch64-linux-gnu/include/alloca.h
```

So, no smoking gun as to why the manual include is necessary in CI, but at least it seems to resolve to the same thing.

---

Digging further, you can detect when it eventually opens the static musl libc:

```
❯ grep 'open.*aarch64.*libc\.a' strace-no-cflags.txt

7447  openat(AT_FDCWD, "/home/ubuntu/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/aarch64-unknown-linux-musl/lib/self-contained/libc.a", O_RDONLY|O_CLOEXEC) = 6
```

(This is the same in both scenarios).

So... this method of cross compiling seems to be doing the following:
- Using clang to target `aarch64-unknown-linux-musl` since clang is "always" a cross compiler (assuming the target triple is enabled when clang is built)
- Uses the llvm linker (`llvm-ar`) to link the copy of static musl `libc.a` that's delivered in the Rust toolchain
- Then, the weird bit: somehow clang (on my Ubuntu VM) is happy to fall back to gcc for sysroot (so any native libs like zlib will be built with clang, llvm-ar and have a GCC sysroot)
- On GitHub CI you have to manually add the sysroot path `/usr/aarch64-linux-gnu/include` to get things to build (for native libs)

So, my guess is this only works as long as the glibc headers that are used by native libs are compatible with the static musl libc.

PR #382, opened 2022-10-13
fix and refactor core and cli

gborough:
Refactored and fixed couple of redundant things like nested inner function and data structure, func params and their types and lifetimes to make everything more uniform and coherent in the codebase

### DIFF
--- a/crates/esthri-cli/src/main.rs
+++ b/crates/esthri-cli/src/main.rs
@@ -44,11 +44,6 @@ async fn log_etag(path: &Path) -> Result<String> {
     Ok(etag)
 }
 
-enum Cli {
-    Esthri(EsthriCli),
-    AwsCompat(AwsCompatCli),
-}
-
 #[derive(Debug, Parser)]
 #[clap(name = "esthri", about = "Simple S3 file transfer utility.")]
 struct EsthriCli {
@@ -503,7 +498,7 @@ async fn dispatch_esthri_cli(cmd: EsthriCommand, s3: &S3Client) -> Result<()> {
         }
 
         DeleteObjects { bucket, keys } => {
-            esthri::delete(s3, bucket, &keys[..]).await?;
+            esthri::delete(s3, &bucket, keys.to_vec()).await?;
         }
 
         #[cfg(not(windows))]
@@ -558,17 +553,6 @@ async fn async_main() -> Result<()> {
         call_real_aws();
     }
 
-    let cli = match aws_compat_mode {
-        false => Cli::Esthri(EsthriCli::parse()),
-        true => {
-            let args = AwsCompatCli::try_parse().map_err(|e| {
-                call_real_aws();
-                anyhow::anyhow!(e)
-            })?;
-            Cli::AwsCompat(args)
-        }
-    };
-
     let region = Region::default();
 
     info!("Starting, using region: {:?}...", region);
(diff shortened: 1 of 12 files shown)


crates/esthri-cli/src/main.rs
```diff
@@ -582,10 +566,19 @@ async fn async_main() -> Result<()> {
     let credentials_provider = DefaultCredentialsProvider::new().unwrap();
     let s3 = S3Client::new_with(http_client, credentials_provider, Region::default());
 
-    match cli {
-        Cli::Esthri(esthri_cli) => dispatch_esthri_cli(esthri_cli.cmd, &s3).await?,
-        Cli::AwsCompat(aws_cli) => dispatch_aws_cli(aws_cli.cmd, &s3).await?,
-    }
+    match aws_compat_mode {
```
silverjam, line 569: Just use an if statement here

crates/esthri/src/blocking.rs
```diff
@@ -19,24 +19,15 @@ use crate::Result;
 use crate::S3PathParam;
 
 #[tokio::main]
-pub async fn head_object<T>(
-    s3: &T,
-    bucket: impl AsRef<str>,
```
silverjam, line 24: impl AsRef<str> is used intentionally throughout the public API to make things more flexible as the call site-- this allows user's to mix String and &str for example

silverjam (changes requested):
the use of `impl AsRef<str>` was an intentional design choice, not really understanding the rationale for changing things back to plain references

crates/esthri-cli/src/main.rs
gborough: > the use of `impl AsRef<str>` was an intentional design choice, not really understanding the rationale for changing things back to plain references

@silverjam 

String already implements AsRef<str> so we can use it freely at call site where both String and &str could be taken, if using String then the caller can simply take a reference, if &str then the function signature is simple instead of using the redundant `as_ref()` which is littered through the code base, like: https://github.com/swift-nav/esthri/blob/d1dc9d9959a7a6a80f61ead288dd449daab50de1/crates/esthri/src/ops/delete.rs#L107
We also have cases where ownership of the str must be taken, and we have it covered with copy-on-write wrapper, like: https://github.com/swift-nav/esthri/blob/03fe7088ad58f9c7d22bce8046f408e24d16ce43/crates/esthri/src/ops/upload.rs#L569, so no issues there

I think earlier code used `impl AsRef<str>` a lot but more recent code started using `&str`, because most of the helper functions that do the heavy lifting take &str anyway, it makes everything simpler. The entire sync.rs module currently only uses &str for example.

I see lot of project getting away from the `impl AsRef<str>` pattern due to some sizedness constraint on some data structures, &str is supposedly more idiomatic. Please correct me if I am wrong on some accounts 😄

crates/esthri-cli/src/main.rs
```diff
@@ -582,9 +566,15 @@ async fn async_main() -> Result<()> {
     let credentials_provider = DefaultCredentialsProvider::new().unwrap();
     let s3 = S3Client::new_with(http_client, credentials_provider, Region::default());
 
-    match cli {
-        Cli::Esthri(esthri_cli) => dispatch_esthri_cli(esthri_cli.cmd, &s3).await?,
-        Cli::AwsCompat(aws_cli) => dispatch_aws_cli(aws_cli.cmd, &s3).await?,
+    if aws_compat_mode.eq(&false) {
```
joelynch, line 569:
```suggestion
    if !aws_compat_mode {
```

crates/esthri/src/blocking.rs
```diff
@@ -151,7 +133,7 @@ pub async fn compute_etag(path: impl AsRef<Path>) -> Result<String> {
 }
 
 #[tokio::main]
-pub async fn delete<T>(s3: &T, bucket: impl AsRef<str>, keys: &[impl AsRef<str>]) -> Result<()>
+pub async fn delete<T>(s3: &T, bucket: &str, keys: Vec<String>) -> Result<()>
```
joelynch, line 136: Why `Vec<String>` over a borrowed type (here and elsewhere)?

joelynch: Seems like you might not want the keys to be consumed when this function is called

joelynch (commented):
Not really a refactor since you are changing the API of this crate in a bunch of places.

crates/esthri/src/blocking.rs
gborough: @joelynch 

> Not really a refactor since you are changing the API of this crate in a bunch of places - we would need to update sitl to support these changes

@joelynch yes this is definitely too much for a single PR so don't worry about it, I will park it here for few days while working on a smaller PR then close it. There are multiple issues with it and hard to address them all at once, since we seems to have some diverging styles written over the years specifically on string handling. I am working on a writeup to address some of the patterns and hopefully you could assist me on this next week if time permits

PR #385, opened 2022-10-13
Remove Windows from Esthri publish CI [WIP] [DEVINFRA-932]

achubbic-snav:
Windows support is removed from Estrhi publishing CI


.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+            needs_musl: true
+
+          - os: macos-11
+            profile: release+mac
+            target: x86_64-apple-darwin
+
+          - os: macos-11
+            profile: release+mac+aarch64
+            target: aarch64-apple-darwin
+
```
silverjam, line 36: We can reduce this matrix to have just windows and mac since Jenkins builds Linux

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
+name: Build macOS, Ubuntu, Windows
+
+on:
+  push:
```
silverjam, line 4: Let's build on `pull_request` events, not tag pushes-- this way we'll see build failures much sooner

achubbic-snav: I am keeping this for testing via Actions tab. Will respond to this convo once changed to pull req events.

.github/workflows/ci.yml
achubbic-snav: I get that that is more efficient in terms of resources used- to me it seems desirable to build everything in the same place, maybe even remove the build from jenkins if we are already doing it here- though I have not looked into the implications of that.

silverjam: Up to you, Linux is the least expensive platform so it's fine to leave it.  However, I don't think we are ready to move off of Jenkins as I don't think our unit/integration tests will run... but it's worth trying at least once.

achubbic-snav: Help me understand the problem: is it the case that there is a machine that runs Jenkins only and that tests are fed to that machine? I seem to recall that being part of our lab walkthrough. <- I am under the impression that if this is true, tests do not presently have a way of running on GHA.

If the above is true, can I create a ticket to look into how to setup a machine dedicated to performing GHA tests?
If the above is false, why would GHA not be capable of running tests?

.github/workflows/ci.yml
achubbic-snav: Changed.

.github/workflows/ci.yml
silverjam: > Help me understand the problem: is it the case that there is a machine that runs Jenkins only and that tests are fed to that machine? I seem to recall that being part of our lab walkthrough. <- I am under the impression that if this is true, tests do not presently have a way of running on GHA.

There are a number of different machines (which are called "executors" or "nodes" in Jenkins) connected to Jenkins, some of them are allocated dynamically (via AWS EC2) and some of them are persistent.  The ones you saw in the lab are persistent, most of the time though we are using the dynamically allocated executors/nodes.  These are preconfigured with credentials that allow them to access our cloud infrastructure (the `esthri-test` bucket in particular).

This mix of dynamic/persistent machines for running tests and compilations is similar on GitHub actions too, except on GitHub Actions the equivalent concept is called a "runner".

> If the above is true, can I create a ticket to look into how to setup a machine dedicated to performing GHA tests?

We have an AWS IAM credential that allows us to access some of our cloud resources from GitHub Actions.  I believe the work would be to adjust this credential so that it can access the `esthri-test`.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: Build macOS, Ubuntu, Windows
 
 on:
-  push:
+  pull_request:
```
silverjam, line 4: tags are not not valid for the `pull_requst` trigger, see documentation: 

this should just be (similar to the `workflow_dispatch` trigger):

```yaml
on:
  pull_request:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Setup cargo make
+        uses: davidB/rust-cargo-make@v1
+
+      - name: Store git tag vars
+        shell: bash
+        run: echo "VERSION=${GITHUB_REF##*/}" >> $GITHUB_ENV
```
adrian-kong, line 51: we dont actually need this / use this anywhere

(we could however, merge `publish.yml` into this `ci.yml` so we won't have to build them twice, i.e. 
https://github.com/swift-nav/libsbp/blob/6cbf963cb708928664d21761c5bf7d2d095ffe8f/.github/workflows/rust.yaml#L100
just a `if: startsWith(github.ref, 'refs/tags/v')` check if its a tag for release stage etc... which in that case will need to run this on all `push:` )

adrian-kong (commented):
looking good so far, think we just need to port the commands from jenkins to run these:

```
cargo make lint
cargo fmt -- --check

cargo make test
```

adrian-kong (commented):
looking good so far, think we just need to port the commands from jenkins to run these:

```
cargo make lint
cargo fmt -- --check

cargo make test
```

.github/workflows/ci.yml
adrian-kong: think this might be helpful
https://github.com/swift-nav/ICBINS/blob/14d8a704e7e42a3d12b2f711b5455df99d9c67e9/.github/workflows/publish.yml#L89-L94

.github/workflows/ci.yml
adrian-kong: in the event we do move towards GHA, it might be useful to build just on `push:` (like libsbp rust.yaml) but check if it has a tag to release so we dont have to build it twice and just upload the ones we built (don't think this affects performance tho since they run simultaneously anyways just reusing code i guess)

.github/workflows/ci.yml
achubbic-snav: Removed.

Paraphrasing our conversation of the (parenthesis) part for @silverjam :
merge could add complications, would rather make it a minor followup ticket

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+
+          - os: macos-11
+            profile: release+mac
+            target: x86_64-apple-darwin
+
+          - os: macos-11
+            profile: release+mac+aarch64
+            target: aarch64-apple-darwin
+
+            - os: windows-2022
```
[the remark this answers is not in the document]
achubbic-snav: Thanks

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+          qemu-user
+
+      - name: Install stable Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: ${{ matrix.target }}
+          profile: minimal
+          default: true
+
```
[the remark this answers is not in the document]
achubbic-snav: Good idea. Done.

.github/workflows/ci.yml
achubbic-snav: Thank you for documentation link. Fixed.

.github/workflows/ci.yml
achubbic-snav: @adrian-kong addressed below

silverjam: IMO we should keep this simple for now, we can optimize double building later

.github/workflows/publish.yml
```diff
@@ -113,7 +110,7 @@ jobs:
         env:
           DL_LINUX: "esthri-${{ env.VERSION }}-x86_64-unknown-linux-musl.tar.gz"
           DL_MAC: "esthri-${{ env.VERSION }}-x86_64-apple-darwin.tar.gz"
-          DL_WIN: "esthri-${{ env.VERSION }}-x86_64-pc-windows-msvc.tar.gz"
+          DL_WIN: "" # Do not publish
```
achubbic-snav, line 113: I think it makes sense to leave this blank rather than removing it, can reviewers think of problems associated with doing so? To me this boils down to the file will not be found either way, so future modifiers will have an easier time tracing windows usage

.github/workflows/ci.yml
achubbic-snav: > looking good so far, think we just need to port the commands from jenkins to run these:
> 
> ```
> cargo make lint
> cargo fmt -- --check
> 
> cargo make test
> ```

Addressed offline. Saving for another ticket.

silverjam: Can we move this back to triggering on `pull_requests`?

.github/workflows/publish.yml
silverjam: 👍

silverjam (commented):
Not sure I'm following why `ci.yaml` is being limited to triggering on tag pushes when the purpose of "ci.yaml" is primarily to validate pull requests-- thinking about what "CI" means ("continuous integration") it follows that these are activities that should happen during the initial proposal and development of pull requests.

achubbic-snav:
> Not sure I'm following why `ci.yaml` is being limited to triggering on tag pushes when the purpose of "ci.yaml" is primarily to validate pull requests-- thinking about what "CI" means ("continuous integration") it follows that these are activities that should happen during the initial proposal and development of pull requests.

I left a note `# TEMPORARY change used to test achubbic` because I wasn't sure how to run action workflows without tag pushing. This will be removed once ticket is in approval state with tests passing.

Alternatively, how can I run a actions workflow for a given commit using `on: pull-request`?

.github/workflows/ci.yml
achubbic-snav: From my previous response: 

I left a note # TEMPORARY change used to test achubbic because I wasn't sure how to run action workflows without tag pushing. This will be removed once ticket is in approval state with tests passing.

Alternatively, how can I run a actions workflow for a given commit using on: pull-request?

silverjam: It triggers when a pull request is opened or updated

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

So all that should be needed is to push an update to the PR to run the new ci.yaml -- you'd like to run the workflow without changes you can force the workflow to re-run through the GitHub web UI or you can simply push an empty commit:

```
git commit --allow-empty -m "trigger ci"
git push
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+    name: Release ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            profile: release+static
+            target: x86_64-unknown-linux-musl
+            needs_musl: true
```
adrian-kong, line 23:
```suggestion
#         - os: ubuntu-latest
#           profile: release+static
#           target: x86_64-unknown-linux-musl
#           needs_musl: true
```

maybe comment this out so we dont have to run an extra build for now until we decide to move jenkins but like @silverjam said, it's negligible, (kinda like having two duplicate tests running same time)

adrian-kong (approved):
besides that, it should be ok,
if all we want is to build windows and not publish.

side note:
think it would be much more simple by just adding field in the windows matrix entirely in `publish.yml ` something rough like...


```diff
on:
  push:
-   tags:
-     - '[0-9]+.[0-9]+.[0-9]+*'
-     - 'v[0-9]+.[0-9]+.[0-9]+*'
```

```diff
          - os: windows-2022
            profile: release+windows
            target: x86_64-pc-windows-msvc
            suffix: .exe
            tar: tar
+           disable: true
```

```diff
      - name: Upload Release
        uses: svenstaro/upload-release-action@v1-release
+       if: ${{ matrix.disable }} == true && startsWith(github.ref, 'refs/tags/v')
        with:
          repo_token: ${{ secrets.GITHUB_TOKEN }}
          file: "*.tar.gz"
          tag: ${{ github.ref }}
          overwrite: true
          file_glob: true
```

```diff
  post_release:
    name: Post release
    needs: [ release ]
    runs-on: ubuntu-18.04
+   if: ${{ matrix.disable }} == true && startsWith(github.ref, 'refs/tags/v')
    steps:
```

to achieve the same results then you wouldn't have to create an entire `ci.yaml` but it's good for future running tests in `ci.yaml`

.github/workflows/ci.yml
achubbic-snav: > It triggers when a pull request is opened or _UPDATED_

Ah excellent. Thank you both.

achubbic-snav: Change in question has been reverted.

silverjam (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
+name: Build macOS, Ubuntu, Windows
+
+on:
+  pull_request:
+
+  workflow_dispatch:
+
+jobs:
+  release:
+    name: Release ${{ matrix.target }}
```
silverjam, line 10: Nit, since this is for CI and not for release (CD) anymore:

```suggestion
    name: Target ${{ matrix.target }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+          #   profile: release+static
+          #   target: x86_64-unknown-linux-musl
+          #   needs_musl: true
+
+          # - os: ubuntu-latest
+          #   profile: release+static+aarch64
+          #   target: aarch64-unknown-linux-musl
+          #   needs_musl: true
+
+          # - os: macos-11
```
silverjam, line 27: macOS is not duplicate for CI, we only build macOS currently on tag pushes

achubbic-snav: I'm not following something. Isn't that true for all of these?

silverjam:
Ship it

PR #395, opened 2022-10-27
more complete sync builder pattern

silverjam:


silverjam:
cc @gborough here's a more complete example of what we were talking through earlier

silverjam:
All the sync tests are failing because of the `derive_builder` crate I used, fixing that should just be a matter of providing defaults for all of the parameters (docs indicate that the `.build()` method only fails a property is unset and their is no default.

silverjam:
cc @gborough fixed the builder so it doesn't crash now, all integration tests are passing now

PR #396, opened 2022-10-31
add sync_down_streaming [DEVINFRA-933]

ziwift:
added `sync_down_streaming` which returns `Stream<Item = Result<Option<String>>` wrapped in `Result`. It is useful to deal with a file as soon as it is downloaded, rather than waiting for the whole batch to be downloaded.
- Some<String> in the Item is the local file path to the downloaded file
- None in the Item means the file is in remote but not synced down (already existed)


silverjam:
cc @notoriaga

crates/esthri/src/ops/sync.rs
```diff
@@ -153,6 +153,58 @@ where
...
+///                   file will be included or excluded. If not supplied, then
+///                   all files will be synced.
+#[logfn(err = "ERROR")]
+pub async fn sync_down_streaming<'a, T>(
+    s3: &'a T,
+    source: &'a S3PathParam,
+    destination: &'a S3PathParam,
+    filters: &'a [GlobFilter],
+    compressed: bool,
+) -> Result<impl Stream<Item = Result<Option<String>>> + 'a>
```
notoriaga, line 175: I'm wondering if instead of `Item = Result<Option<String>>` maybe something like `Item = Result<SyncResult>`

```rust
pub enum SyncResult  {
    Synced(String),
    Cached(String),
}
impl SyncResult {
    pub fn into_path(self) -> String { ... }
}
```

because for my particular use case I still need the names of the already downloaded files. I suppose I could just walk the directory before syncing and chain that stream with the one returned here. Could go either way. Also don't really like the name/variant names but can't think of anything better

ziwift: ended up creating a struct, wdyt? @notoriaga 
```
pub enum Synced {
    /// newly created file after sync
    Created,
    /// already existed file
    Existed,
}

pub struct SyncedFile {
    synced: Synced,
    src_path: String,
    dest_path: String,
}
```

crates/esthri/src/ops/sync.rs
```diff
@@ -894,6 +943,96 @@ where
...
+            directory.into(),
+            etag_stream,
+            opts,
+        );
+
+        Box::pin(
+            sync_tasks
+                .buffer_unordered(task_count)
+                .take_while(|_task| future::ready(true)), // always go through the whole stream
+        )
```
notoriaga, line 1032: I think just this should work 

```suggestion
        Box::pin(sync_tasks.buffer_unordered(task_count))
```

notoriaga (approved):
Thanks!

PR #398, opened 2022-11-04
Builder Pattern Refactor [DEVINFRA-965]

gborough:
The purpose of this PR:

1. Consolidation of multiple variants of the same "verbs" into unified interface
2. Implement builder patterns for all optional parameters for all "verbs" wherever possible
3. Breaking up some crates into more manageable and extensible smaller units
4. Cleaning up few redundant data structures and fix lint

### DIFF
--- a/crates/esthri-cli/Cargo.toml
+++ b/crates/esthri-cli/Cargo.toml
@@ -45,7 +45,7 @@ mime_guess = "2.0"
 once_cell = "1.7"
 sanitize-filename = "0.4"
 serde = "1"
-clap = { version = "4.0.18", features = ["derive", "env"] }
+clap = { version = "3.2.18 ", features = ["derive", "env"] }
 tokio = { version = "1.6", features = ["rt-multi-thread", "signal", "sync"] }
 tokio-util = { version = "0.7.4", features = ["compat", "codec"] }
 warp = "0.3"
--- a/crates/esthri-cli/Cargo.toml
+++ b/crates/esthri-cli/Cargo.toml
@@ -33,6 +33,7 @@ async-compression = { version = "0.3", features = ["gzip", "tokio"] }
 async_zip = "0.0.7"
 async-stream = "0.3"
 bytes = "1.0"
+derive_builder = "0.11.2"
 env_logger = "0.9"
 futures = { version = "0.3", default-features = false }
 glob = "0.3"
@@ -44,7 +45,7 @@ mime_guess = "2.0"
 once_cell = "1.7"
 sanitize-filename = "0.4"
 serde = "1"
-clap = { version = "3.2.22", features = ["derive"] }
+clap = { version = "4.0.18", features = ["derive", "env"] }
 tokio = { version = "1.6", features = ["rt-multi-thread", "signal", "sync"] }
 tokio-util = { version = "0.7.4", features = ["compat", "codec"] }
 warp = "0.3"
--- a/crates/esthri-cli/Cargo.toml
+++ b/crates/esthri-cli/Cargo.toml
@@ -45,7 +45,7 @@ mime_guess = "2.0"
 once_cell = "1.7"
 sanitize-filename = "0.4"
 serde = "1"
-clap = { version = "3.2.18 ", features = ["derive", "env"] }
+clap = { version = "3.2.22", features = ["derive", "env"] }
 tokio = { version = "1.6", features = ["rt-multi-thread", "signal", "sync"] }
 tokio-util = { version = "0.7.4", features = ["compat", "codec"] }
 warp = "0.3"
--- a/crates/esthri-cli/tests/integration/http_server.rs
+++ b/crates/esthri-cli/tests/integration/http_server.rs
@@ -6,8 +6,7 @@ use std::{fs, io};
 use tokio::runtime::Runtime;
 use warp::http::status::StatusCode;
 
-use esthri::upload;
-use esthri::{blocking, upload_compressed};
+use esthri::{blocking, opts::*, upload, Config};
 use esthri_server::esthri_filter;
 use esthri_test::{validate_key_hash_pairs, DateTime, KeyHashPair};
 
@@ -19,12 +18,14 @@ async fn test_fetch_object() {
     let bucket = esthri_test::TEST_BUCKET;
 
     let s3_key = "test_file.txt".to_owned();
+    let opts = GenericOptParamsBuilder::default().build().unwrap();
 
     let res = upload(
         s3client.as_ref(),
         esthri_test::TEST_BUCKET,
         &s3_key,
         &filename,
+        &opts,
     )
     .await;
     assert!(res.is_ok());
@@ -54,12 +55,14 @@ async fn test_allowed_prefixes() {
     // Nominal happy path, a prefix is allowed, so access to a file is allowed.
 
     let an_allowed_key = "allowed_prefix/test_file.txt".to_owned();
+    let opts = GenericOptParamsBuilder::default().build().unwrap();
 
     let res = upload(
         s3client.as_ref(),
         esthri_test::TEST_BUCKET,
         &an_allowed_key,
         &filename,
+        &opts,
     )
     .await;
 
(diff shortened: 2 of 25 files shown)


john-michaelburke:
nit. I would say generally it is good practice to have even a vague description of the content of the PR in the title and the convention we use is square brackets for adding a ticket tag.
```Refactor crates into smaller units[DEVINFRA-965]```

Also if there are some clear cut boundaries in the multiple changes for this PR, it may be better to break them out into smaller PRs. In contrast, if you had 5 small changes that all occurred in a few files, in my opinion, that would be a case when it would be reasonable to throw them all into one PR.

File: crates/esthri/src/blocking.rs [hunk 1]
```diff
@@ -36,52 +32,12 @@ pub async fn upload<T>(
     bucket: impl AsRef<str>,
     key: impl AsRef<str>,
     file: impl AsRef<Path>,
+    opts: &GenericOptParams,
```
joelynch, line 35: I think I would prefer to have a different OptParams type for each verb.  Right now you can do stuff like

```rust
let opts = GenericOptParamsBuilder::default().transparent_compression(true).build().unwrap();
let res = head_object(..., &opts);
```

This firstly doesn't make any sense but maybe more importantly doesn't help the user of the API understand which options are available for this verb.  As a user of the API, I would want to look at the definition of `GenericOptParams` here to understand which options are available for `head_object`.

joelynch (commented):
Would prefer per verb opt structs

crates/esthri-cli/tests/integration/http_server.rs
```diff
@@ -373,11 +386,18 @@ fn test_fetch_archive_with_compressed_files() {
     let filename = esthri_test::test_data("index.html");
     let s3client = esthri_test::get_s3client();
     let s3_key = format!("{}/{}", prefix, "index_compressed.html");
-    let res = blocking::upload_compressed(
+    let opts = GenericOptParamsBuilder::default().build().unwrap();
+    let opts_compressed = GenericOptParamsBuilder::default()
```
joelynch, line 390: nit: I think I am more of a fan of this pattern where you can access the builder via a `::builder()` function on the original struct.

```rust
let opts = GenericOptParams::builder()
    .transparent_compression(true)
    .build();
```

joelynch:
Also to get it to link to jira the title needs to be of the form

```
Builder Pattern Refactor [DEVINFRA-965]
```

crates/esthri-cli/Cargo.toml
```diff
@@ -44,7 +45,7 @@ mime_guess = "2.0"
 once_cell = "1.7"
 sanitize-filename = "0.4"
 serde = "1"
-clap = { version = "3.2.22", features = ["derive"] }
+clap = { version = "3.2.18 ", features = ["derive", "env"] }
```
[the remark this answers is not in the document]
gborough: It was set to latest version but build failed due to nativetls msrv, accidentally rolled back to the wrong version. Thanks for the pick up

crates/esthri/src/blocking.rs
gborough: Initially I have gone down the path of different struct for different verb using respective param holders, however problems quickly arose, seeing that the lib crates has lots of interwoven verbs calling each other, e.g. sync function internally calling upload and download, or copy calling both upload and download etc, while these verbs are mostly consolidated into a single function, they end up with lots of shared parameters if they were to work with a single call

I think it's definitely a tradeoff is to have a shared struct where the members are well-known s3 related options like compression and delete. The other option is to have another go at diff struct for diff verb but need another round of refactoring to decouple these interwoven function. What do you think?

crates/esthri/src/ops/upload.rs
```diff
@@ -116,103 +119,14 @@ where
...
-    let compressed = true;
-    upload_file_helper(
-        s3,
-        bucket.as_ref(),
-        key.as_ref(),
-        file.as_ref(),
-        compressed,
-        Config::global().storage_class(),
+        opts.transparent_compression,
+        opts.storage_class
```
joelynch, line 128: You can set default values in the `derive_builder` which might be an alternative to this.

crates/esthri/src/blocking.rs
joelynch: If you implement `From` for easily transforming structs for some verbs into others does that help make it less convoluted?

If we had flattened or nested builders it might help out too but looks like this isn't implemented yet. 
 https://github.com/colin-kiegel/rust-derive-builder/issues/254

silverjam: I was thinking this same thing when I first looked over these changes: the need for different builder types will help with the overall clarity of the API. I agree that `From` is probably the best way to solve the interdependence issue, it'll require some boilerplate, but the gain in API clarity is worth it IMO.

crates/esthri/src/lib.rs
```diff
@@ -328,3 +319,77 @@ where
...
     Ok(listing)
 }
+
+pub mod opts {
+    use derive_builder::Builder;
+    use glob::Pattern;
+
+    use crate::rusoto::*;
+    #[derive(Debug, Copy, Clone, Builder)]
+    pub struct AwsCopyOptParams {
```
joelynch, line 329: Should this be `EsthriCopyOptParams` for consistency?

crates/esthri/src/lib.rs
```diff
@@ -328,3 +319,77 @@ where
...
+        #[builder(default = "Some(S3StorageClass::Standard)")]
+        pub storage_class: Option<S3StorageClass>,
```
joelynch, line 331: This shouldn't be an option anymore if it has a default

crates/esthri/src/ops/copy.rs
```diff
@@ -12,42 +12,36 @@
...
 {
     match source {
         S3PathParam::Bucket { bucket, key } => match destination {
             S3PathParam::Local { path } => {
-                if transparent_compression {
-                    download_with_transparent_decompression(s3, bucket, key, path).await
-                } else {
-                    download(s3, bucket, key, path).await
-                }
+                download(s3, bucket, key, path, EsthriGetOptParams::from(opts)).await
```
joelynch, line 36: nit: `opts.into()` is less verbose - any reason not to go with it?

crates/esthri/src/lib.rs
```diff
@@ -328,3 +319,77 @@ where
...
+        fn from(opt: AwsCopyOptParams) -> Self {
+            Self {
+                storage_class: opt.storage_class,
+                transparent_compression: opt.transparent_compression,
+            }
+        }
+    }
+
+    #[derive(Debug, Copy, Clone, Builder)]
+    pub struct EsthriGetOptParams {
```
joelynch, line 363: This is way better - I can immediately see what options I have for esthri GET.

joelynch (approved):
Final few nits but this looks really good to me now!

crates/esthri/src/lib.rs
gborough: This is for the aws compat mode, the esthri mode doesn't have the copy option so I thought I make a distinction here

crates/esthri/src/ops/copy.rs
gborough: sure

crates/esthri/src/lib.rs
silverjam: Removing the Option wrapper should also simplify the usage of this field.

silverjam (approved):
One remaining minor improvement for AwsCopyOptParams, otherwise this looks great

crates/esthri/src/lib.rs
gborough: The reason Option wrapper is used for storageclass is because of this line:
https://github.com/swift-nav/esthri/blob/e29f684b5a618bab6d0a23682ec931444100f28d/crates/esthri/src/ops/upload.rs#L129

If storageclass is not set then we are taking the value from the env variable. Using Option wrapper would allow it to use unwrap_or_else() so that we would potentially write less boilerplate sin the upload and cli crates for matching different storageclass value

silverjam: Missed that, thanks

silverjam:
:shipit:

PR #399, opened 2022-11-08
Presigned URLs [DEVINFRA-933]

joelynch:
Added some functions for generating pre-signed urls for S3 Get, Put, Delete and Multipart Put.  Since it can be tricky to use pre-signed urls properly, which I found out during testing, I've also added in helpers for using pre-signed urls using the `reqwest` library.

TODO
- [ ] handle transparent compression for the multipart put
- [ ] test transparent compression
- [x] fix openssl missing error on CICD
- [ ] Add `presign` command to CLI


gborough (approved):

crates/esthri/src/presign/upload.rs
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+use std::{os::unix::prelude::MetadataExt, path::Path, time::Duration};
```
adrian-kong, line 13:
```suggestion
use std::{path::Path, time::Duration};
```
think this unix specific - can delete this if we use len() - (*not actually sure the difference tho the docs sounds like they do exactly the same thing*)

crates/esthri/src/presign/upload.rs
```diff
@@ -0,0 +1,86 @@
...
+/// Helper to download a file using a presigned URL, setting appropriate
+/// headers.
+pub async fn upload_file_presigned(
+    client: &Client,
+    presigned_url: &str,
+    filepath: &Path,
+    opts: EsthriPutOptParams,
+) -> Result<()> {
+    let file = file_maybe_compressed(filepath, &opts).await?;
+    let file_size = file.metadata().await?.size();
```
adrian-kong, line 66:
```suggestion
    let file_size = file.metadata().await?.len();
```
should this be len() instead?

crates/esthri/Cargo.toml
```diff
@@ -44,6 +44,8 @@ tokio-util = { version = "0.7", features = ["io"] }
 walkdir = "2"
 strum = "0.24.1"
 strum_macros = "0.24.3"
+reqwest = { version="0.11.12", features=["stream", "json"] }
```
adrian-kong, line 47: could we slap this into test dependency,
presigned urls dont need https i believe

so u can probably also do
`default-features = false`

crates/esthri/src/presign/upload.rs
john-michaelburke: Looks like they return the same thing :shrug: 
https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.size

crates/esthri/src/presign/upload.rs
joelynch: Good pick up, changed

crates/esthri/Cargo.toml
joelynch: reqwest is used in the presigned url client code - I could make this a specific feature maybe `presigned_url_client`.

joelynch:
Maybe this work should go under some specific features

silverjam (approved):

PR #411, opened 2022-11-22
Presign get command

adrian-kong:


crates/esthri-cli/src/cli_opts.rs
```diff
@@ -228,3 +228,13 @@ pub struct EsthriServeParams {
...
 }
+
+#[derive(Args, Debug, Clone)]
+pub struct EsthriPresignParams {
+    /// The bucket to target (example: my-bucket)
+    #[clap(long)]
+    pub bucket: String,
+    /// The key to target (example: a/key/name.bin)
+    #[clap(long)]
+    pub key: String,
```
joelynch, line 239: Do we want to add a "method" parameter here?  Will differ from `aws s3` but I think could be useful.

joelynch: I'll do this

joelynch: we can work on this later

PR #419, opened 2022-12-06
add granular credential provider options [DEVINFRA-1038]

gborough:
For better security and granular level of control, Esthri should be recommended to use a specific credential provider via an environment variable, only to fall back to the default credential provider if variable not set should the users choose to

### DIFF
--- a/crates/esthri-cli/src/cli_utils.rs
+++ b/crates/esthri-cli/src/cli_utils.rs
@@ -6,7 +6,7 @@ use std::{
     time::Duration,
 };
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use clap::ArgMatches;
 use esthri::{rusoto::*, GlobFilter};
 use glob::Pattern;
@@ -139,7 +139,7 @@ pub fn setup_upload_termination_handler(s3: S3Client) {
     });
 }
 
-pub fn setup_s3client_with_cred_provider() -> S3Client {
+pub fn setup_s3client_with_cred_provider() -> Result<S3Client> {
     let mut hyper_builder = Client::builder();
     hyper_builder.pool_idle_timeout(Duration::from_secs(20));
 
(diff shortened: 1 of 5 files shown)


crates/esthri-cli/src/cli_utils.rs
```diff
@@ -136,3 +138,40 @@ pub fn setup_upload_termination_handler(s3: S3Client) {
...
+            }
+            "container" => {
+                let credentials_provider = ContainerProvider::new();
+                S3Client::new_with(http_client, credentials_provider, Region::default())
+            }
+            "instance_metadata" => {
+                let credentials_provider = InstanceMetadataProvider::new();
+                S3Client::new_with(http_client, credentials_provider, Region::default())
+            }
+            _ => {
```
silverjam, line 167: As a quality of life improvement we should allow using `ESTHRI_CREDENTIAL_PROVIDER= esthri .,.` to mean the default credential provider, this will allow users to easily override an environment variable without have to delete it from the environment.

crates/esthri-cli/src/cli_utils.rs
```diff
@@ -136,3 +138,40 @@ pub fn setup_upload_termination_handler(s3: S3Client) {
...
+                println!("unset or unsupported credential provider environment variable, program aborting");
+                std::process::exit(1)
```
silverjam, line 169: `main` for esthri already handles `Result` types, we should return a `Result` instead of forcing a process exit

crates/esthri-cli/src/cli_utils.rs
```diff
@@ -136,3 +138,67 @@ pub fn setup_upload_termination_handler(s3: S3Client) {
...
+            }
+            "instance_metadata" => {
+                let credentials_provider = InstanceMetadataProvider::new();
+                Ok(S3Client::new_with(
+                    http_client,
+                    credentials_provider,
+                    Region::default(),
+                ))
+            }
+            "esthri" => {
```
silverjam, line 183:
```suggestion
            "" => {
```

Is there any downside to making empty and unset do the same thing?  It's a pretty common convention in environment variable handling to do this (a convention that we've used in other Swift tools too).

gborough: @silverjam No downside that I can see, happy to align with other tool's convention

silverjam: 👍

silverjam (approved):

PR #438, opened 2023-02-01
rust: bump MSRV to 1.60.0

silverjam:
Some recent dependency updates are failing because they bumping their MSRV beyond ours, particularly clap is bumping there's to 1.60.0-- we supported 1.56.1 simply because that was the oldest one that worked (without any additional effort) so proposing we bump to stay inline with clap (and any other important dependencies).


pcrumley:
oh shoot, i just made a PR that does the same thing.. oops. Anyway like i said in the other PR i think it might be worth it to take the opportunity to set the MSRV to 1.66.1 to just update past the cargo CVE listed here: https://blog.rust-lang.org/2023/01/10/cve-2022-46176.html. don't feel strongly about it though

pcrumley (approved):
make sure we bump esthri by a major version when we release this

silverjam:
> oh shoot, i just made a PR that does the same thing.. oops. 

Sorry, should've ping you before I jumped on this.

> Anyway like i said in the other PR i think it might be worth it to take the opportunity to set the MSRV to 1.66.1 to just update past the cargo CVE listed here: https://blog.rust-lang.org/2023/01/10/cve-2022-46176.html. don't feel strongly about it though

I think this'll cause issues for SITL until we update it to work with a stable toolchain, so I'd rather revisit this once SITL is updated-- don't want to be in a situation where we can't bring an Esthri bugfix into SITL because they have conflicting MSRV values.

image

PR #439, opened 2023-02-01
bump rust-msrv and upgrade clap to 4.1

pcrumley:
this PR will require a major version update of Esthri, as it moves the minimal supported rust version to 1.60.0 from 1.56.1

FWIW-- given the CVE detailed here: https://blog.rust-lang.org/2023/01/10/cve-2022-46176.html it might make sense to use this opportunity to upgrade msrv to 1.66.1


pcrumley:
closing in favor of https://github.com/swift-nav/esthri/pull/438

PR #558, opened 2023-09-05
migrate to aws_sdk

ziwift:
- migrated from rusoto to the official aws sdk for all aws interactions
- removed the retry logic, because it is already implemented in the sdk's s3 `Client` struct

breaking changes:
1. use sdk's Client to replace the rusoto S3 Client 
2. use the native sdk `StorageClass` instead of the one defined in this crate
3. in `presign` module, these functions now 1) are async and 2) takes aws_sdk_s3 `Client` as input to replace rusoto `AwsCredentials` and `Region`:
- `presign_delete`
- `presign_get` 
- `setup_presigned_multipart_upload`
- `complete_presigned_multipart_upload`
- `abort_presigned_multipart_upload`
- `presign_put`


crates/esthri-internals/Cargo.toml
pcrumley, line 1: Kinda just skimming this, but why didn't you remove the rusoto dependencies? Does this PR completely remove rusoto from esthri?

crates/esthri-test/src/lib.rs
```diff
@@ -24,7 +23,7 @@ pub struct TestGlobal {
 static TEST_GLOBAL: Lazy<Mutex<TestGlobal>> =
     Lazy::new(|| Mutex::new(TestGlobal { s3client: None }));
 
-const TEST_GLOBAL_LOCK_FAILED: &str = "locking global test data failed";
+// const TEST_GLOBAL_LOCK_FAILED: &str = "locking global test data failed";
```
silverjam, line 26: Remove this if it isn't used anymore

silverjam (approved):
We should bump to version 11 for the next release

silverjam:
We also need to resolve the conflicts with master/main -- @ljbade made some improvements to the error messages that we need to port.

PR #579, opened 2023-10-10
support creation of aws clients in a blocking context

pcrumley:
Esthri supports calling some function in a blocking context using the `#[tokio::main]` macro. Unfortunately with the move to the aws_sdk crate, the new aws client can only be created in an async function, sort of completely destroying the point of the blocking.rs module.

Here I added the ability to create a default client in a blocking context, so one can use blocking.rs modules functions in an e2e scenario in a different crate.

In addition, I want to be able to specify the client's region, so I added that to the s3 client initiator.

### DIFF
--- a/crates/esthri/src/lib.rs
+++ b/crates/esthri/src/lib.rs
@@ -76,17 +76,35 @@ pub enum AwsCredProvider {
 }
 
 pub async fn init_default_s3client() -> Client {
-    init_s3client(AwsCredProvider::DefaultProvider).await
+    init_default_s3client_with_region(None::<&str>).await
+}
+
+pub async fn init_default_s3client_with_region(region: Option<impl AsRef<str>>) -> Client {
+    init_s3client_with_region(AwsCredProvider::DefaultProvider, region).await
 }
 
 pub async fn init_s3client(provider: AwsCredProvider) -> Client {
+    init_s3client_with_region(provider, None::<&str>).await
+}
+
+pub async fn init_s3client_with_region(
+    provider: AwsCredProvider,
+    region: Option<impl AsRef<str>>,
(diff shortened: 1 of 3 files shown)


crates/esthri/src/lib.rs
```diff
@@ -76,17 +76,35 @@ pub enum AwsCredProvider {
...
     let retry_config = aws_config::retry::RetryConfig::standard()
         .with_initial_backoff(Duration::from_millis(500))
         .with_max_attempts(5);
     let https_connector = new_https_connector();
     let smithy_connector = hyper_ext::Adapter::builder().build(https_connector);
 
-    let sdk_config = aws_config::load_from_env().await;
+    let sdk_config = if let Some(region) = region {
+        aws_config::from_env()
+            .region(Region::new(region.as_ref().to_owned()))
```
adrian-kong, line 102: Can't you pass this as env?
```
$ export AWS_DEFAULT_REGION=us-west-2
```

silverjam (approved):

adrian-kong (approved):
lgtm

crates/esthri/src/lib.rs
pcrumley: hm. so you are correct, but in this case I want to build a client inside of a cli with a given region to match the location of the binary we are trying to download. I feel like messing with the user's env vars is not a good solution as it is a weird side-effect.

If you disagree with this approach, I suppose i can either build the client inside of the cli code, or use something like 
https://crates.io/crates/temp-env

pcrumley: I think i will add some comments to the code that it may be better to set the env var depending on your use case

PR #583, opened 2023-10-12
bump async_zip to latest

pcrumley:
This PR handles the upgrading of `async_zip` to the latest version. They did some work around supporting multiple async runtimes which we really don't benefit from here because we are just using tokio.

Anyway, this gets us unstuck from async_zip 0.7

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,19 +85,15 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.3.12"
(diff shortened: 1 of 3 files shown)


pcrumley:
pinging @silverjam because he 👍🏻  the original dependabot commit.

note for semver, this did change some type signatures for functions, but can be considered a patch release for esthri because none of those functions are public.

silverjam (approved):

PR #621, opened 2024-01-03
Upgrades to aws-sdk 1.0

pcrumley:
This PR updates Esthri to AWS SDK 1.0. 

Changes:
 - Add a private wrapper type with type erasure because AWS removed the stream trait implementation on S3 ByteStream
 - Wanted to remove the deprecated aws-smithy-types-conversion crate so I convert the s3 DateTime -> std's SystemTime -> chrono::DateTime<UTC>
 - The rest are just pretty standard updates to handle breaking changes in the SDK API, e.g. infallible things became fallible, things started to return options, and some of the handling of the http client were spun off into the smith-runtime-api crate.

### DIFF
--- a/crates/esthri/src/aws_sdk.rs
+++ b/crates/esthri/src/aws_sdk.rs
@@ -146,8 +146,8 @@ pub struct GetObjectResponse {
     pub size: i64,
     pub part: i64,
 }
-/// The AWS SDK removed future stream implementation of a Byte Stream Wrapper,
-/// so we implement it here
+/// The AWS SDK removed future stream implementation of a ByteStream,
+/// so we implement it here on a wrapper struct that holds the ByteStream.
 struct FutureStreamByteStreamWrapper(ByteStream);
 
 impl FutureStreamByteStreamWrapper {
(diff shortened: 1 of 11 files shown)


crates/esthri/src/aws_sdk.rs
```diff
@@ -129,10 +146,28 @@ pub struct GetObjectResponse {
     pub size: i64,
     pub part: i64,
 }
+/// The AWS SDK removed future stream implementation of a Byte Stream Wrapper,
```
notoriaga, line 149: Kind of odd given that ByteStream has both `poll_next` and `size_hint`, wonder what the reasoning is behind that

crates/esthri/src/aws_sdk.rs
```diff
@@ -129,10 +146,28 @@ pub struct GetObjectResponse {
...
+}
+
+impl Stream for FutureStreamByteStreamWrapper {
+    type Item = Result<Bytes>;
 
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        Pin::new(&mut self.0)
+            .poll_next(cx)
+            .map(|opt| opt.map(|res| res.map_err(|e| Error::ByteStreamError(e.to_string()))))
+    }
```
notoriaga, line 166: nit: you could also impl `Stream::size_hint` with `ByteStream::size_hint`

notoriaga (approved):

PR #669, opened 2024-03-08
S3object meta

kdesjard:
capture more of the object's metadata into S3Object, specifically:

storage_class
size
last_modified

### DIFF
--- a/crates/esthri/src/types.rs
+++ b/crates/esthri/src/types.rs
@@ -17,6 +17,8 @@ use std::{
     result::Result as StdResult,
 };
 
+use aws_sdk_s3::primitives::DateTime;
+use aws_sdk_s3::types::ObjectStorageClass;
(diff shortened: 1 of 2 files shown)


crates/esthri/src/types.rs
```diff
@@ -152,6 +154,9 @@ impl S3Listing {
 pub struct S3Object {
```
pcrumley, line 154:
```suggestion
#[non_exhaustive]
pub struct S3Object {
```

pcrumley (commented):
@kdesjard 

Thank you for your interest is esthri and for submitting this PR. Somehow I missed this alert so I apologize for not responding sooner.

I think it is a good idea to add non_exhaustive to this struct so then if you want to add additional attributes in the future (e.g. owner etc), we won't have to push a major version in the update.

PR #704, opened 2026-03-20
Set BucketOwnerFullControl ACL on multipart uploads

reimerix:
## Summary
- `create_multipart_upload` was missing `.acl(ObjectCannedAcl::BucketOwnerFullControl)` that `singlepart_upload` and `empty_upload` already set
- This caused cross-account uploaded objects (via `esthri sync`) to be inaccessible to the bucket owner account, resulting in 403 errors on `HeadObject`/`GetObject`
- Discovered while migrating Starling CI from Jenkins to GitHub Actions — Jenkins used `s3Upload` with explicit `BucketOwnerFullControl`, while `esthri sync` silently omitted it for files above the multipart threshold

## Test plan
- [ ] Existing integration test `cli_test.rs` already tests `--acl bucket-owner-full-control`
- [ ] Verify cross-account upload via `esthri sync` sets correct ACL on multipart objects

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/ci.yml
```diff
@@ -24,11 +24,11 @@ jobs:
             target: aarch64-unknown-linux-musl
             needs_musl: true
 
-          - os: macos-11
+          - os: macos-14
```
jbangelo: We probably should make this `macos-latest` to avoid breaking the CI again when this version gets removed from underneath us.

jbangelo (approved):
Just one note, but that can be addressed in a different PR.

